PR #29, opened 2025-05-20
Chapter table stubs

Josh-Beck:
I took the liberty of adding stub titles and tables to each chapter, as well as a first pass try at a brief description for most. Figured it would be a good jumping off point once the real info was getting added. Feel free to use or ignore!


jmanico (approved):

jmanico:
Thank you kindly for this PR!

PR #44, opened 2025-05-30
Add Appendix E "Example Tools and Frameworks"

Josh-Beck:
Here is the suggested addition of a tooling appendix to provide developers quick guidance on how to complete a specific AISVS requirement. 

A few options for how this could be numbered:
* Each AISVS chapter and section will receive it's own Appendix section `AE.chapter.section`. This is currently the format for this PR.
* Each chapter could have a section, with tooling listed and applicable sections listed.
* Each type of tooling could have it's own line item, perhaps with logic breaks in the table for different tooling types like access control and IaC.


jmanico (commented):
A few notes about this:

1) Lets make this vendor neutral and remove the NVidia reference and other vendor references throughout the standard eventually
2) This is such a good idea, I think the appendix is the building blocks for all developers and the main developer resource
3) With this in mind, I am likely to suggest we remove the D/V sections for all other requirements - and essentially make them all "v"
4) Them this appendix is a list of all controls developers need to implement
5) And then we map all verifiable requirements for testers to the controls in this section

What do you think, Josh?

jmanico (approved):

mbhatt1:
Thanks for doing this buddy

PR #49, opened 2025-06-05
Add initial document generation setup

ottosulin:
This has been shamelessly copied and adapted from ASVS repository.

With 100% this does not work. However only way to see all the things that break is the first merge this to main and then begin to fix.


jmanico:
Merged! Let's do it!

PR #56, opened 2025-06-25
Edit c05-Access-Control-and-Identity

thornshadow99:
Made minor edits to requirements and body text in C05. 

NOTES: 

**5.2.1**: do we use "allow-lists" (used here) or "allowlists" (used elsewhere in AISVS)?

**5.4.2**: does the prepositional phrase "with policy inheritance" apply to all three clauses? I presume it does.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,5 +1,0 @@
-{
-    "githubPullRequests.ignoredPullRequestBranches": [
-        "main"
-    ]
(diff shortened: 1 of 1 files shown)


.vscode/settings.json
jmanico, line 1: Can you remove this file from your PR?

jmanico (approved):

PR #86, opened 2025-10-20
chat set validation suggestion

jmanico:
chat set validation suggestion from @theJonMccoy


jmanico (commented):
Work from @theJonMccoy

Issue #144, opened 2026-03-13
Proposed controls: Deterministic safety layers, tool-usage verification, hash-chain audit, and context window flooding


mattijsmoens:
Hi team,

After a thorough review of AISVS v1.0, I wanted to suggest a few controls that address attack vectors and defense patterns I've encountered in production. Happy to submit PRs for any of these if there's interest.

1. Deterministic safety layer requirements (C2 / C11)

The standard references "content classifiers" (C2.5.1) and "adaptive detection models" (C2.8.2) but doesn't distinguish between ML based and deterministic safety mechanisms. When the safety layer itself is an ML model, it inherits the same adversarial vulnerabilities as the model it protects prompt injection, adversarial perturbation, jailbreaking.

Proposed control under C2 or C11: Verify that safety critical input/output filtering includes at least one deterministic enforcement layer (e.g., compiled regex pattern matching, cryptographic hash verification, or explicit allow/block lists) that cannot be bypassed through adversarial ML techniques.

This doesn't replace ML based classifiers, it ensures there's always a hard floor that can't be gradient attacked.

2. Tool usage verification for hallucination detection (C7.2)

C7.2 covers confidence scoring and retrieval based grounding, but misses a practical detection signal: whether the model actually invoked verification tools before making factual claims. A model asserting "the current temperature is 72F" without having called a weather API is a strong hallucination indicator regardless of confidence score.

Proposed control under C7.2: Verify that the system tracks tool/function invocation history within a request chain and flags high confidence factual assertions that were not preceded by relevant verification tool usage.

3. Hash chain integrity for audit logs (C9.4)

C9.4.3 recommends "append only/WORM/immutable log store" for tamper evidence, which requires specialized storage infrastructure. Cryptographic hash chains (each audit entry includes the hash of the previous entry) provide portable tamper evidence that works with any storage backend and can be independently verified by any party.

Proposed addition to C9.4.3: Verify that tamper evident audit mechanisms include cryptographic chaining (e.g., each record includes the hash of the prior record) or equivalent integrity guarantees that can be independently verified without relying solely on storage layer immutability.

4. Context window flooding defense (C2.1)

The standard addresses prompt injection broadly, but doesn't specifically cover context window flooding. Where an attacker provides extremely long inputs designed to push system instructions out of the model's effective attention window, causing the model to lose its safety instructions.

[...]

PR #145, opened 2026-03-13
Add cryptographic hash chaining as tamper evidence option for C9.4.3

mattijsmoens:
Related to #144

Extends C9.4.3 to include cryptographic hash chaining as a tamper evidence mechanism alongside append only/WORM/immutable log store. Hash chains (each record includes the hash of the prior record) provide portable tamper evidence that works with any storage backend, can be independently verified by any party, and detects tampering even if the storage layer is compromised. Reference implementation: https://github.com/mattijsmoens/sovereign-shield (CoreSafety module, SHA-256 hash chains for audit entries)


1.0/en/0x10-C09-Orchestration-and-Agentic-Action.md
```diff
@@ -55,7 +55,7 @@ Make every action attributable and every mutation detectable.
...
 | **9.4.2** | **Verify that** agent-initiated actions are cryptographically bound to the execution chain (chain ID) and are signed and timestamped for non-repudiation and traceability. | 2 | D/V |
-| **9.4.3** | **Verify that** audit logs are tamper-evident (append-only/WORM/immutable log store) and include sufficient context to reconstruct who/what acted, initiating user identifier, delegation scope, authorization decision (policy/version), tool parameters, approvals (where applicable), and outcomes. | 2 | D/V |
+| **9.4.3** | **Verify that** audit logs are tamper-evident (via append-only/WORM/immutable log store, cryptographic hash chaining where each record includes the hash of the prior record, or equivalent integrity guarantees that can be independently verified) and include sufficient context to reconstruct who/what acted, initiating user identifier, delegation scope, authorization decision (policy/version), tool parameters, approvals (where applicable), and outcomes. | 2 | D/V |
```
ottosulin: I think "tamper-evident" summarizes our intent of integrity guarantees. I'm not sure if we need to document such implementation detail.

Where the implementation details are not specific to AI, I would leave it to be looked up from other sources and keep it high-level; we just set this as a requirement for AI systems on certain level of security.

mattijsmoens: That's a fair point. "Tamper-evident" does cover the intent well, and I agree the standard should stay implementation-agnostic. I could simplify the wording. Would something like "tamper-evident (via append-only storage, cryptographic chaining, or equivalent integrity guarantees that can be independently verified)" work? That keeps it high-level while acknowledging there's more than one valid approach.

PR #147, opened 2026-03-13
Add content based type detection and filename sanitization controls u…

mattijsmoens:
Related to #144

Adds two controls to C2.7 (Multi Modal Input Validation) addressing AI-specific file type risks:

- C2.7.6: Content-based file type detection to prevent attackers from disguising malicious payloads as valid input modalities (e.g., a script with a .png extension bypassing image-only safety filters and reaching the model pipeline unvalidated).

- C2.7.7: Filename sanitization against null byte injection, path traversal, and double extension attacks that could manipulate how AI pipelines route, store, or process uploaded files (e.g., "prompt.jpg.exe" or "../../model_config.json").

Both are deterministic, zero-cost checks that sit at the boundary between user input and model ingestion, before any modality-specific AI safety filters are applied.

Reference implementation: https://github.com/mattijsmoens/sovereign-shield (MultiModalFilter module)

### DIFF
--- a/1.0/en/0x10-C02-User-Input-Validation.md
+++ b/1.0/en/0x10-C02-User-Input-Validation.md
@@ -97,7 +97,6 @@ AI systems should include robust validation for non-textual inputs (images, audi
 | **2.7.6** | **Verify that** file type validation uses content-based detection (magic bytes/file headers) in addition to declared MIME types and file extensions, so that type-spoofed files (e.g., a script disguised as an image) cannot bypass modality-specific AI safety filters and reach the model pipeline unvalidated, and that mismatches between declared and detected type are rejected as potential type spoofing. | 1 | D/V |
 | **2.7.7** | **Verify that** filenames submitted to AI processing pipelines are sanitized against null byte injection, path traversal sequences, and double extension attacks (e.g., "input.jpg.exe") that could manipulate how the pipeline routes, stores, or interprets uploaded files before model ingestion. | 1 | D/V |
 
-
 ---
 
 ## C2.8 Real-Time Adaptive Threat Detection


1.0/en/0x10-C02-User-Input-Validation.md
```diff
@@ -94,6 +94,8 @@ AI systems should include robust validation for non-textual inputs (images, audi
 | **2.7.3** | **Verify that** image/audio inputs are checked for adversarial perturbations or known attack patterns, and detections trigger gating (block or degrade capabilities) before model use. | 2 | D/V |
 | **2.7.4** | **Verify that** multi-modal input validation failures trigger detailed logging including all input modalities, validation results, threat scores, and trace metadata (source, tool or MCP server, agent ID, session as applicable), and generate alerts for investigation. | 3 | D/V |
 | **2.7.5** | **Verify that** cross-modal attack detection identifies coordinated attacks spanning multiple input types (e.g., steganographic payloads in images combined with prompt injection in text) with correlation rules and alert generation, and that confirmed detections are blocked or require HITL (human-in-the-loop) approval. | 3 | D/V |
+| **2.7.6** | **Verify that** file type validation uses content based detection (magic bytes/file headers) in addition to declared MIME types and file extensions, and that mismatches between declared and detected type are rejected as potential type spoofing. | 1 | D/V |
```
ottosulin: While I agree this should be done, I have the concern as in #145 - we should try to keep to our scope of AI without starting to replicate e.g. ASVS contents when going to details.

@jmanico this PR is a good example of something we should discuss and probably clarify in Preface / Using AISVS: scope of the project and relation to other standards and frameworks.

We do currently include controls that are e.g. about runtime environment which are applicable also to non-AI applications. But replicating same control implementation details from e.g. ASVS may not be useful. On the other hand, I think we do kind of expect & want AISVS L2 systems to be roughly on same level of security as ASVS L2 - same controls are needed.

mattijsmoens: Good point on scope. These controls exist in ASVS for web apps, but in the AI context they sit at the boundary between user input and model ingestion the risk isn't just RCE from a malicious upload, it's that a spoofed file type could bypass modality specific safety filters (e.g., a script disguised as an image bypassing image only validation). I can adjust the wording to emphasize the AI specific angle if that helps.

jmanico: That works for me. If you can frame this as a specific AISVS control that is not covered in ASVS yet, I'm all for it.

But Otto is right, we do want to avoid repeating the core of ASVS. I'm happy to make exceptions where appropriate.

ottosulin: Now checked against ASVSv5, I think that is [covered](https://github.com/OWASP/ASVS/blob/master/5.0/en/0x14-V5-File-Handling.md) already. I would here rather interpret ASVS broadly to cover similar use cases to prevent AISVS and ASVS unnecessary overlaps.

That being said I agree with Jim, if in the world of file handling there is an _AI specific implementation that would not exist without AI_ very interested to discuss and include it.

mattijsmoens: You're right, V5.2.2 and V5.3.2 cover the core file validation mechanisms sufficiently. I am closing this one. Thanks for the thorough review.

PR #178, opened 2026-03-16
Add safety violation feedback loop security under C11.9

mattijsmoens:
Related to #144

Adds a control for securing the feedback loop when safety violations are used as training data. Blocked prompt injections, caught hallucinations, and filtered policy violations are high-value training data, but attackers could intentionally trigger violations to poison the improvement pipeline. This control requires integrity verification, poisoning detection, and human review gates.

Reference implementation: https://github.com/mattijsmoens/sovereign-shield (AdaptiveShield module - learns from reported attacks with false positive validation)


mattijsmoens:
Thank you very much for the kind words and yes it is indeed one of the most overlooked gaps. I am glad we agree on that.

Issue #158, opened 2026-03-15
Update C09 - Semantic corruption propagation in agent chains


almogbhl:
As suggested in #151 , once it will be merged, I will open a PR with the verification to fill that gap

Issue #157, opened 2026-03-15
Update C09 - System prompt integrity at load time


almogbhl:
As suggested in #151 , once it will be merged, I will open a PR with the verification to fill that gap

PR #183, opened 2026-03-16
Update c09 agentic additions

almogbhl:
**Issues:**
#157 
#158 

**Additions:**
1. prompt integrity at load time
2. semantic validations

Full suggestions discussed #151

### DIFF
--- a/1.0/en/0x10-C09-Orchestration-and-Agentic-Action.md
+++ b/1.0/en/0x10-C09-Orchestration-and-Agentic-Action.md
@@ -69,6 +69,7 @@ Protect agent-to-agent and agent-to-tool communications from hijacking, injectio
 | **9.5.1** | **Verify that** agent-to-agent and agent-to-tool channels enforce mutual authentication and encryption using current recommended protocols (e.g., TLS 1.3 or later) with strong certificate/token validation. | 1 | D/V |
 | **9.5.2** | **Verify that** all messages are strictly schema-validated; unknown fields, malformed payloads, and oversized frames are rejected. | 1 | D/V |
 | **9.5.3** | **Verify that** message integrity covers the full payload including tool parameters, and that replay protections (nonces/sequence numbers/timestamp windows) are enforced. | 2 | D/V |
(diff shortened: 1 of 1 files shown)


1.0/en/0x10-C09-Orchestration-and-Agentic-Action.md
```diff
@@ -94,6 +95,7 @@ Prevent “technically authorized but unintended” actions by binding execution
 | **9.7.1** | **Verify that** pre-execution gates evaluate proposed actions and parameters against hard policy constraints (deny rules, data handling constraints, allow-lists, side-effect budgets) and block execution on any violation. | 1 | D/V |
 | **9.7.2** | **Verify that** high-impact actions require explicit user intent confirmation that is integrity-protected and bound to the exact action parameters (and expires quickly) to prevent stale or substituted approvals. | 2 | D/V |
 | **9.7.3** | **Verify that** post-condition checks confirm the intended outcome and detect unintended side effects; any mismatch triggers containment (and compensating actions where supported). | 2 | V |
+| **9.7.4** | **Verify that** prompt templates and agent policy configurations are integrity-verified at load time against their approved versions (e.g., via hashes or signatures). | 2 | D/V |
```
ottosulin, line 98: I began thinking that in most apps the prompts & templates come from the same repo so no sense to do this -> maybe add in the end "... if retrieved from remote sources"?

Also, I'm not aware of this existing today in the common frameworks like langchain or Google ADK (please correct if I am wrong here). If it would be, I would say this fits L2 however because this requires customization of existing frameworks to go that far today I would bump this to L3.

almogbhl: L3 - agreed.

And for the same reason we sign Docker images even though the Dockerfile is in our own repo -> stuff can get tampered with between "git push" and "running in production." This control says: do the same for your prompts. Hash them at build, verify at load. 

wdyt?

jmanico: I'm going to merge this and force it to level 3. Please re-open or add a new PR if you think I'm off base!

PR #191, opened 2026-03-17
C9.8: expand multi-agent swarm risk controls with four new requirements

RicoKomenda:
Add 9.8.3–9.8.6 to address gaps in the multi-agent domain isolation and swarm risk controls section:

- 9.8.3 (L1, D/V): System-level cap on simultaneously active agent instances enforced by the orchestration runtime, preventing spawning loops and unbounded swarm growth / resource exhaustion.

- 9.8.4 (L2, D/V): Lateral movement prevention - each agent is technically restricted to its own memory namespace, context window, and credentials and cannot read or modify peer agent state.

- 9.8.5 (L2, D/V): Swarm-level aggregate action rate limits across all agents (API calls, file writes, network requests per time window) to prevent coordinated bursts causing unintended external impact.

- 9.8.6 (L3, D/V): Swarm-level emergency shutdown capability that halts all active agent instances simultaneously, revokes tool and MCP access, and blocks new spawning within a pre-defined response time (complements the per-model shutdown in C3.3.4/3.3.5).

### DIFF
--- a/1.0/en/0x10-C09-Orchestration-and-Agentic-Action.md
+++ b/1.0/en/0x10-C09-Orchestration-and-Agentic-Action.md
@@ -107,10 +107,10 @@ Reduce cross-domain interference and emergent unsafe collective behavior.
 | :--: | --- | :---: | :--: |
 | **9.8.1** | **Verify that** agents in different tenants, security domains, or environments (dev/test/prod) run in isolated runtimes and network segments, with default-deny controls that prevent cross-domain discovery and calls. | 1 | D/V |
 | **9.8.2** | **Verify that** runtime monitoring detects unsafe emergent behavior (oscillation, deadlocks, uncontrolled broadcast, abnormal call graphs) and automatically applies corrective actions (throttle, isolate, terminate). | 3 | D/V |
-| **9.8.3** | **Verify that** a system-level cap on simultaneously active agent instances is enforced by the orchestration runtime, so that spawning loops or recursive sub-agent creation cannot cause unbounded swarm growth or resource exhaustion. | 1 | D/V |
-| **9.8.4** | **Verify that** each agent can only access its own memory namespace, context window, and credentials, and is technically prevented from reading or modifying peer agent state, preventing lateral movement across agents within the same swarm. | 2 | D/V |
-| **9.8.5** | **Verify that** swarm-level aggregate action rate limits (e.g., total external API calls, file writes, or network requests per time window across all agents) are enforced to prevent coordinated bursts that cause unintended impact on external systems. | 2 | D/V |
-| **9.8.6** | **Verify that** a swarm-level emergency shutdown capability exists that halts all active agent instances simultaneously, revokes their tool and MCP access, and prevents new agent spawning, with the shutdown completable within a pre-defined response time. | 3 | D/V |
+| **9.8.3** | **Verify that** each agent is restricted to its own memory namespace and is technically prevented from reading or modifying peer agent state, preventing unauthorized cross-agent access within the same swarm. | 2 | D/V |
+| **9.8.4** | **Verify that** each agent operates with an isolated context window and dedicated credentials scoped to its role, preventing peer agents from accessing or influencing another agent's context or credential scope to prevent unauthorized cross-agent access within the same swarm. | 3 | D/V |
+| **9.8.5** | **Verify that** swarm-level aggregate action rate limits (e.g., total external API calls, file writes, or network requests per time window across all agents) are enforced to prevent bursts that cause denial-of-service or abuse of external systems. | 3 | D/V |
+| **9.8.6** | **Verify that** a swarm-level shutdown capability exists that can halt all active agent instances or selected problematic instances in an organized fashion and prevents new agent spawning, with shutdown completable within a pre-defined response time. | 3 | D/V |
 
 ---
 


1.0/en/0x10-C09-Orchestration-and-Agentic-Action.md
```diff
@@ -107,13 +105,13 @@ Reduce cross-domain interference and emergent unsafe collective behavior.
 | :--: | --- | :---: | :--: |
 | **9.8.1** | **Verify that** agents in different tenants, security domains, or environments (dev/test/prod) run in isolated runtimes and network segments, with default-deny controls that prevent cross-domain discovery and calls. | 1 | D/V |
 | **9.8.2** | **Verify that** runtime monitoring detects unsafe emergent behavior (oscillation, deadlocks, uncontrolled broadcast, abnormal call graphs) and automatically applies corrective actions (throttle, isolate, terminate). | 3 | D/V |
+| **9.8.3** | **Verify that** a system-level cap on simultaneously active agent instances is enforced by the orchestration runtime, so that spawning loops or recursive sub-agent creation cannot cause unbounded swarm growth or resource exhaustion. | 1 | D/V |
```
ottosulin: Well spawning loops are a real risk in recursive sub-agent architectures. Two concerns:

- Overlap with 9.1.1: The existing requirement already covers "max fan-out/concurrency" as per-execution budgets. What makes this distinct is the system-wide cap across all concurrent executions. Consider making that distinction explicit in the wording, e.g., "...independent of per-execution budgets.". Or propose a change to that requirement.
- Feasibility: Afaik no major framework (LangGraph, CrewAI, Google ADK, OpenAI Agents SDK) provides a system-wide active instance cap. They offer per-execution limits (recursion_limit, max_turns, max_iter) which are already covered by 9.1.1. A true system-wide cap requires infrastructure-level enforcement (process pools, K8s replica limits, etc).

1.0/en/0x10-C09-Orchestration-and-Agentic-Action.md
```diff
@@ -107,13 +105,13 @@ Reduce cross-domain interference and emergent unsafe collective behavior.
...
+| **9.8.4** | **Verify that** each agent can only access its own memory namespace, context window, and credentials, and is technically prevented from reading or modifying peer agent state, preventing lateral movement across agents within the same swarm. | 2 | D/V |
```
ottosulin: The intent is solid. Two issues:

- Overlap with 8.5.5 (L2): That requirement already covers multi-agent memory namespace isolation enforced through access control. The new value here is context window and credential isolation, which 8.5.5 does not address.
- Level: Framework support is very thin. CrewAI offers memory.scope() for soft namespace scoping, but it's a view filter, not a hard access control boundary. Google ADK and LangGraph explicitly share session.state across parallel agents with no isolation. Context window isolation and per-agent credential separation exist in zero frameworks, both require fully custom engineering. Suggest either:
  * Split: memory namespace isolation at L2 (partially supported, and overlaps with 8.5.5) + context window and credential isolation at L3 (no framework support), or
  * Move the whole thing to L3 and add a cross-reference to 8.5.5 for the memory piece.

Minor language thing: "lateral movement" is a network security term. "Unauthorized cross-agent access" may fit this AI context better.

1.0/en/0x10-C09-Orchestration-and-Agentic-Action.md
```diff
@@ -107,13 +105,13 @@ Reduce cross-domain interference and emergent unsafe collective behavior.
...
+| **9.8.5** | **Verify that** swarm-level aggregate action rate limits (e.g., total external API calls, file writes, or network requests per time window across all agents) are enforced to prevent coordinated bursts that cause unintended impact on external systems. | 2 | D/V |
```
ottosulin: This fills a genuine gap. Existing rate limits in 2.6.x and 9.1.x are all per-entity or per-execution. The scenario where 50 individually-compliant agents collectively overwhelm an external API is not covered anywhere else in the standard.

No framework provides aggregate swarm-level rate limiting, but the implementation pattern is well-understood (e.g. shared Redis-backed token bucket) and reasonable for security-conscious teams. I would consider moving to L3 because requires today custom engineering, and we have mitigating L1-L2 controls included.

Minor wording suggestion:s "unintended impact on external systems" -> "denial-of-service or abuse of external systems" for precision, and remove the word "coordinated".

1.0/en/0x10-C09-Orchestration-and-Agentic-Action.md
```diff
@@ -107,13 +105,13 @@ Reduce cross-domain interference and emergent unsafe collective behavior.
...
+| **9.8.6** | **Verify that** a swarm-level emergency shutdown capability exists that halts all active agent instances simultaneously, revokes their tool and MCP access, and prevents new agent spawning, with the shutdown completable within a pre-defined response time. | 3 | D/V |
```
ottosulin: Revoking accesses isn't necessarily simple, may require out-of-band access with separate creds to disable credentials if needed. I would keep the requirement focused on being able to shutdown runaway system in an organized fashion without resorting to "yanking the cable".

I would remove the unnecessary word "emergency".

Also, I think the capability should not be limited to shut down all or none; ideally shutdown can be applied to all or to selected problematic agent instances.

1.0/en/0x10-C09-Orchestration-and-Agentic-Action.md
RicoKomenda: So you would say to (for now) remove this req, @ottosulin?
Maybe I need to put up an issue at the frameworks 🤣

PR #193, opened 2026-03-17
C1.6: add Bias Detection & Fairness Assurance section

RicoKomenda:
The chapter is titled 'Training Data Governance & Bias Management' but contained no bias-specific requirements. This adds C1.6 with five new requirements to close that gap:

- 1.6.1 (L1, D/V): Baseline bias measurement across protected attribute groups (demographic parity, equalized odds, disparate impact ratio) recorded as part of the dataset approval record.

- 1.6.2 (L2, D/V): Datasets exceeding defined bias thresholds are blocked from training until remediation is applied, re-evaluated, and approved by an authorized person.

- 1.6.3 (L2, D): Debiasing/rebalancing techniques applied when thresholds are exceeded, with measured effect logged before training proceeds.

- 1.6.4 (L3, V): Intersectional bias evaluation covering combinations of protected attributes (e.g., race × gender) tracked separately from single-attribute metrics.

- 1.6.5 (L3, D/V): Bias metrics re-evaluated after any significant dataset change; regressions trigger a remediation workflow before training proceeds.

Aligns with EU AI Act Article 10 (data governance obligations) and NIST AI RMF MAP/MEASURE functions.

### DIFF
--- a/.github/workflows/config/local-custom.txt
+++ b/.github/workflows/config/local-custom.txt
@@ -100,3 +100,6 @@ mbhatt
 cciprofm
 thornshadow
 Rego
+debiasing
+rebalancing
+intersectional


jmanico:
I'm going to close this out for now, as described above we only want bias requirements that are security related. Is that acceptable?

PR #194, opened 2026-03-17
C10.1: expand MCP component integrity with version pinning, manifest signing, and namespace isolation

RicoKomenda:
Add 10.1.2–10.1.4 to address gaps in MCP supply chain hygiene:

- 10.1.2 (L1, D/V): Version pinning and server allowlisting - MCP server dependencies pinned to exact versions; only allowlisted identifiers (name + version + registry) accepted in production; unpinned or unlisted servers rejected at load time.

- 10.1.3 (L2, D/V): Cryptographic signature verification for tool manifests - MCP capability descriptors must be signed by the server and verified by the client before being used to configure tool invocations; unsigned or tampered manifests are rejected.

- 10.1.4 (L2, D/V): Tool namespace isolation - tool names are scoped to their originating MCP server namespace and uniqueness is enforced per server by the runtime, preventing tool-name squatting attacks where a malicious server shadows a legitimate tool name.

### DIFF
--- a/1.0/en/0x10-C10-MCP-Security.md
+++ b/1.0/en/0x10-C10-MCP-Security.md
@@ -12,7 +12,7 @@ Ensure secure discovery, authentication, authorization, transport, and use of MC
 | :--: | --- | :---: | :--: |
 | **10.1.1** | **Verify that** MCP server and client components are obtained only from trusted sources and verified using signatures, checksums, or secure package metadata, rejecting tampered or unsigned builds. | 1 | D/V |
 | **10.1.2** | **Verify that** MCP client and server configurations do not contain plaintext secrets (API keys, tokens, client secrets) and that credentials are injected or resolved at runtime rather than stored in configuration files, environment variables, or source code. | 1 | D/V |
-| **10.1.3** | **Verify that** packages and plugins loaded by MCP servers are sourced only from allowlisted registries and constrained to approved version ranges (e.g., pinned to major version), so that dependency confusion or substitution attacks cannot introduce unauthorized code that exposes malicious tools to the AI agent runtime. | 1 | D/V |
+| **10.1.3** | **Verify that** only allowlisted MCP server identifiers (name, version, and registry) are permitted in production and that the runtime rejects connections to unlisted or unregistered servers at load time. | 1 | D/V |
 
 ---
 


1.0/en/0x10-C10-MCP-Security.md
```diff
@@ -12,6 +12,8 @@ Ensure secure discovery, authentication, authorization, transport, and use of MC
 | :--: | --- | :---: | :--: |
 | **10.1.1** | **Verify that** MCP server and client components are obtained only from trusted sources and verified using signatures, checksums, or secure package metadata, rejecting tampered or unsigned builds. | 1 | D/V |
 | **10.1.2** | **Verify that** MCP client and server configurations do not contain plaintext secrets (API keys, tokens, client secrets) and that credentials are injected or resolved at runtime rather than stored in configuration files, environment variables, or source code. | 1 | D/V |
+| **10.1.3** | **Verify that** MCP server dependencies are pinned to exact versions and that only allowlisted server identifiers (name, version, and registry) are permitted in production; version-unpinned or unlisted servers are rejected at load time. | 1 | D/V |
+| **10.1.4** | **Verify that** tool names are scoped to their originating MCP server namespace and that the runtime enforces uniqueness per server, preventing tool-name squatting where a malicious or misconfigured server registers a tool with the same name as a legitimate one. | 2 | D/V |
```
ottosulin: This addresses a real attack vector (tool-name squatting/shadowing), but I believe it's already covered by existing C10.4 controls:

- 10.4.2 (L2) requires tool schemas to be "validated for authenticity and integrity using signatures." If the client verifies a tool's schema against the originating server's identity, a malicious server's tool can't pass authenticity checks as a legitimate server's tool -- which is the core defense against shadowing.
- 10.4.5 (L2) requires clients to "maintain a hash or versioned snapshot of tool definitions" and re-approve on changes. A new server registering a duplicate tool name would trigger this re-approval gate.

Together, these two requirements already prevent tool-name squatting if implemented properly. Namespace scoping is one implementation technique for achieving what 10.4.2 + 10.4.5 already require as outcomes. Please elaborate if you disagree!

1.0/en/0x10-C10-MCP-Security.md
```diff
@@ -12,6 +12,8 @@ Ensure secure discovery, authentication, authorization, transport, and use of MC
 | :--: | --- | :---: | :--: |
 | **10.1.1** | **Verify that** MCP server and client components are obtained only from trusted sources and verified using signatures, checksums, or secure package metadata, rejecting tampered or unsigned builds. | 1 | D/V |
 | **10.1.2** | **Verify that** MCP client and server configurations do not contain plaintext secrets (API keys, tokens, client secrets) and that credentials are injected or resolved at runtime rather than stored in configuration files, environment variables, or source code. | 1 | D/V |
+| **10.1.3** | **Verify that** MCP server dependencies are pinned to exact versions and that only allowlisted server identifiers (name, version, and registry) are permitted in production; version-unpinned or unlisted servers are rejected at load time. | 1 | D/V |
```
ottosulin: I think here you are referring to e.g. Python package dependencies?

If yes, I'm slightly averse to including generic security best practices here. My smoke test for these being: _Will this control make sense also without AI?_  Like also Jim has commented elsewhere, open to exceptions to this rule if the concern is particularly valid for AI use cases.

1.0/en/0x10-C10-MCP-Security.md
RicoKomenda: Yeah, that's a little bit obsolet then. I'll go with you on this one! :)

1.0/en/0x10-C10-MCP-Security.md
RicoKomenda: Not only python package dependencies but all dependencies in general for a MCP server (which could also refrain to third-party tool use done through a MCP server/gateway

RicoKomenda:
@ottosulin & @jmanico 
I worked the comments in. We have now one new req - please review, if that is sufficient and plausbile for you

1.0/en/0x10-C10-MCP-Security.md
ottosulin: Thanks for clarifying, Rico. I think the updated requirement (as revised in your second commit) actually conflates two distinct concerns:

- Generic dependency version pinning (the MCP server's own packages/libraries) which is already covered by C6.3.1 ("all package managers enforce version pinning via lockfiles") and C6.3.2 (immutable digests for container references). It applies equally to any software, not just MCP servers. Including it here would duplicate C6. And now thinking I think we should consider removing those, but that's another discussion to be had later.
- MCP server allowlisting (only approved server identifiers permitted in production, unlisted rejected at load time): this is MCP-specific. 10.1.1 covers obtaining components from trusted sources with signature verification, but doesn't explicitly require a runtime allowlist that blocks unlisted servers at load time. That's a legitimate gap.

My suggestion: strip the generic dependency-pinning half (since C6 already covers it) and keep the MCP-specific server allowlisting as a focused requirement. Something like:

> Verify that only allowlisted MCP server identifiers (name, version, and registry) are permitted in production and that the runtime rejects connections to unlisted or unregistered servers at load time.

This way it complements 10.1.1 (which covers how you verify integrity) by adding which servers are permitted in the first place - an MCP-specific allow/deny enforcement point.

What do you think?

RicoKomenda: Really like that. I did a new commit with your suggestion. :)

PR #209, opened 2026-03-18
C3.2: add safety re-validation after model compression control (3.2.5)

RicoKomenda:
Quantization, pruning, and distillation alter weight distributions in ways that disproportionately degrade safety-critical parameters, since alignment guardrails do not reside in the highest-magnitude weights that compression algorithms prioritize preserving.

Two distinct threat models:
1. Incidental regression: standard INT4/AWQ/GPTQ quantization silently degrades refusal behavior (demonstrated across 24 settings, arXiv:2502.15799).
2. Adversarial supply chain: attackers embed dormant backdoors in full-precision models that only activate post-quantization, making safety evaluations on the base model meaningless for the deployed artifact (arXiv:2405.18137).

New 3.2.5 (L2, D/V): treats every compressed model as a distinct artifact requiring independent safety and alignment validation with separately signed evaluation artifacts.

### DIFF
--- a/1.0/en/0x10-C03-Model-Lifecycle-Management.md
+++ b/1.0/en/0x10-C03-Model-Lifecycle-Management.md
@@ -31,7 +31,7 @@ Models must pass defined security and safety validations before deployment.
 | **3.2.2** | **Verify that** security testing covers agent workflows, tool and MCP integrations, RAG and memory interactions, multimodal inputs, and guardrails (safety models or detection services) using a versioned evaluation harness. | 2 | D/V |
 | **3.2.3** | **Verify that** all model changes (deployment, configuration, retirement) generate immutable audit records including a timestamp, an authenticated actor identity, a change type, and before/after states, with trace metadata (environment and consuming services/agents) and a model identifier (version/digest/signature). | 2 | V |
 | **3.2.4** | **Verify that** validation failures automatically block model deployment unless an explicit override approval from pre-designated authorized personnel with documented business justifications. | 3 | D/V |
-| **3.2.5** | **Verify that** models subjected to post-training quantization, pruning, or distillation are re-evaluated against the full safety and alignment test suite on the compressed artifact before deployment, and that the resulting safety evaluation artifacts are separately signed and stored. | 2 | D/V |
+| **3.2.5** | **Verify that** models subjected to post-training quantization, pruning, or distillation are re-evaluated against the same safety and alignment test suite on the compressed artifact before deployment, and that evaluation results are retained as distinct records linked to the compressed artifact's version or digest. | 2 | D/V |
 
 ---
 


1.0/en/0x10-C03-Model-Lifecycle-Management.md
```diff
@@ -31,6 +31,7 @@ Models must pass defined security and safety validations before deployment.
 | **3.2.2** | **Verify that** security testing covers agent workflows, tool and MCP integrations, RAG and memory interactions, multimodal inputs, and guardrails (safety models or detection services) using a versioned evaluation harness. | 2 | D/V |
 | **3.2.3** | **Verify that** all model changes (deployment, configuration, retirement) generate immutable audit records including a timestamp, an authenticated actor identity, a change type, and before/after states, with trace metadata (environment and consuming services/agents) and a model identifier (version/digest/signature). | 2 | V |
 | **3.2.4** | **Verify that** validation failures automatically block model deployment unless an explicit override approval from pre-designated authorized personnel with documented business justifications. | 3 | D/V |
+| **3.2.5** | **Verify that** models subjected to post-training quantization, pruning, or distillation are re-evaluated against the full safety and alignment test suite on the compressed artifact before deployment, and that the resulting safety evaluation artifacts are separately signed and stored. | 2 | D/V |
```
ottosulin: Before my lengthy explanation of request for changes - this is a good proposal, proven by research and also makes sense from first principles.

Some feedback:

1. Overlap with 3.2.1 The existing 3.2.1 already requires safety evaluations "before deployment" which, if interpreted strictly, should cover the compressed artifact since that is what gets deployed. The main value of 3.2.5 is making the compression triggers explicit testing explicitly rather than relying on correct interpretation.

2. "Separately signed and stored" clause The signing aspect overlaps with 3.1.2 which already requires cryptographic signing of model artifacts. Could we simplify the requirement to focus on the core obligation (re-evaluation of compressed artifacts) and let the existing signing requirements in 3.1.2 cover the artifact integrity? 

3. Minor language thing: "full" qualifier "Full safety and alignment test suite" is slightly ambiguous. If the intent is that the same test suite used for the uncompressed model must be re-run (rather than a subset), I would prefer "the same safety and alignment test suite used for the uncompressed model" for clarity. Alternatively, just drop "full" since 3.2.1 already defines what the test suite must include.

Overall this is a good addition. The level (L2) is appropriate given that the re-evaluation is straightforward for organizations already complying with 3.2.1.

Suggested rewording:

> Verify that models subjected to post-training quantization, pruning, or distillation are re-evaluated against the same safety and alignment test suite on the compressed artifact before deployment, and that evaluation results are retained as distinct records linked to the compressed artifact's version or digest.

This removes the "separately signed" part (already covered by 3.1.2) and replaces "full safety and alignment test suite" with unambiguous "same safety and alignment test suite".

ottosulin (changes requested):

PR #210, opened 2026-03-18
C5.5: add inference KV-cache isolation control (5.5.5)

RicoKomenda:
LLM inference engines cache Key-Value attention matrices to avoid recomputing shared prefixes. In multi-tenant deployments, cross-tenant prefix reuse creates a timing side-channel: by measuring time-to-first-token latency, an attacker distinguishes cache hits from misses and reconstructs another user's prompt token-by-token.

Demonstrated at NDSS 2025 (PromptPeek) with 92.3% average token recovery accuracy using ~234 probes per token against real serving infrastructure. A separate inversion attack (arXiv:2508.09442) shows KV-cache matrices can also be algebraically inverted to recover input tokens when direct cache access is obtained.

Existing 5.5.3 covers memory spaces and embedding stores generically but does not address inference KV-cache, which operates at the serving layer and requires explicit partitioning by principal identity.

New 5.5.5 (L2, D): requires KV-cache partitioning by authenticated session/tenant identity and prohibits cross-principal prefix sharing.

### DIFF
--- a/.github/workflows/config/local-custom.txt
+++ b/.github/workflows/config/local-custom.txt
@@ -101,5 +101,5 @@ cciprofm
 thornshadow
 Rego
 unvalidated
-
 Cylance
+NDSS


1.0/en/0x10-C05-Access-Control-and-Identity.md
```diff
@@ -70,6 +70,7 @@ Ensure logical and cryptographic isolation between tenants in shared AI infrastr
 | **5.5.2** | **Verify that** every API request includes an authenticated tenant identifier that is cryptographically validated against session context and user entitlements. | 1 | D/V |
 | **5.5.3** | **Verify that** memory spaces, embedding stores, cache entries, and temporary files are namespace-segregated per tenant with secure purging on tenant deletion or session termination. | 2 | D/V |
 | **5.5.4** | **Verify that** encryption keys are unique per tenant with customer-managed key (CMK) support and cryptographic isolation between tenant data stores. | 3 | D |
+| **5.5.5** | **Verify that** inference-time KV-cache entries are partitioned by authenticated session or tenant identity and that shared prefix caching across distinct principals is disabled or isolated to prevent timing-based prompt reconstruction attacks. | 2 | D |
```
ottosulin: Good addition! The PromptPeek research and the broader KV-cache side-channel research justifies a dedicated control here. A few thoughts:

- Overlap with 5.5.3: The existing 5.5.3 mentions "cache entries" generically. It would be helpful to clarify the relationship, perhaps by adding a brief note in the PR description or a parenthetical in 5.5.3 that it covers data-layer caches (embedding caches, result caches) while 5.5.5 specifically targets inference-layer KV-cache prefix sharing. Without this, auditors may wonder if 5.5.3 already covers this.

- Wording suggestion: The current text is ok, but I'd suggest a small tweak:

> "Verify that inference-time KV-cache entries are partitioned by authenticated session or tenant identity and that automatic prefix caching does not share cached prefixes across distinct security principals, to prevent timing-based prompt reconstruction attacks."

Rationale: "automatic prefix caching" is the specific term used by vLLM/SGLang (APC), and "security principals" is slightly more precise than just "principals" in a security standard context. The "disabled or isolated" phrasing is fine, but "does not share...across" is more direct about the security property we want.

Level 2 seems ok: vLLM has cache salting, SGLang has multi-tenant prefix tagging, and even the OpenAI API scopes prefix caching per-org. So I'd say is implementable with current tooling.

PR #219, opened 2026-03-20
C3.7: add fine-tuning pipeline authorization and reward model integrity

RicoKomenda:
Four genuine gaps not covered by 3.1.2/3.4.3/3.4.4/11.1.4 or any other existing requirement:

- 3.7.1 fine-tuning trigger authorization: distinct from 3.4.3 (env isolation) and 3.4.4 (data validation) - governs who may initiate a run and requires separation of duty with immutable pre-execution audit log.

- 3.7.2 reward model as signed artifact: distinct from 3.1.2 (deployment- admission signing) - reward models are consumed during training, not deployed, so deployment-time integrity checks do not cover them.

- 3.7.3 reward hacking detection: distinct from 11.9.5 (feedback pipeline integrity) and 11.1.4 (RLHF reproducibility) - requires a held-out human-preference probe set and divergence gate, not a feedback loop check.

- 3.7.4 multi-stage checkpoint signing: distinct from 3.1.2 (final artifact), 3.2.3 (change audit), 3.3.3 (full rollback) - targets intra-pipeline intermediate stage artifacts to bound blast radius of a stage compromise.

### DIFF
--- a/1.0/en/0x10-C03-Model-Lifecycle-Management.md
+++ b/1.0/en/0x10-C03-Model-Lifecycle-Management.md
@@ -91,10 +91,10 @@ Fine-tuning pipelines are high-privilege operations that can alter deployed mode
 
 | # | Description | Level | Role |
 |:--------:|---------------------------------------------------------------------------------------------------------------------|:---:|:---:|
-| **3.7.1** | **Verify that** initiating a fine-tuning run requires authenticated authorization from a designated approver separate from the person requesting the run, with the approval identity, timestamp, and approved training configuration recorded in an immutable audit log before pipeline execution begins. | 2 | D/V |
-| **3.7.2** | **Verify that** the reward model used in any RLHF fine-tuning stage is treated as a versioned, signed artifact subject to the same integrity verification and validation gates required of policy model artifacts before it is used in a training run. | 2 | D/V |
-| **3.7.3** | **Verify that** reward hacking is evaluated after each RLHF training stage by measuring the policy model's reward scores on a held-out probe set and comparing them to human preference scores on the same set, with the run blocked from promotion if divergence exceeds a defined threshold. | 3 | D/V |
-| **3.7.4** | **Verify that** in multi-stage fine-tuning pipelines, each stage's output checkpoint is cryptographically signed and recorded as a distinct artifact in the model registry before the next stage begins, enabling isolation and rollback of any compromised intermediate stage. | 3 | D/V |
+| **3.7.1** | **Verify that** initiating a fine-tuning or retraining run requires authorization from a person who did not request the run (separation of duties). | 3 | D/V |
+| **3.7.2** | **Verify that** reward models used in RLHF fine-tuning are versioned, cryptographically signed, and integrity-verified before use in a training run. | 2 | D/V |
+| **3.7.3** | **Verify that** RLHF training stages include automated detection of reward hacking or reward model over-optimization (e.g., held-out human-preference probe sets, divergence thresholds, or KL penalty monitoring), with the run blocked from promotion if detection thresholds are exceeded. | 3 | D/V |
+| **3.7.4** | **Verify that** in multi-stage fine-tuning pipelines, each stage's output is integrity-verified before the next stage consumes it, and intermediate checkpoints are registered as distinct artifacts enabling per-stage rollback. | 3 | D/V |
 
 ---
 


1.0/en/0x10-C03-Model-Lifecycle-Management.md
```diff
@@ -85,6 +85,19 @@ Hosted and provider-managed models may change behavior without notice. These con
 
 ---
 
+## C3.7 Fine-Tuning Pipeline Authorization & Reward Model Integrity
+
+Fine-tuning pipelines are high-privilege operations that can alter deployed model behavior at scale. Multi-stage pipelines compound this risk because a compromise at any intermediate stage produces a subtly altered artifact that subsequent stages accept. Reward models used in RLHF are ML artifacts subject to tampering yet often treated as static infrastructure rather than versioned, validated components.
+
+| # | Description | Level | Role |
+|:--------:|---------------------------------------------------------------------------------------------------------------------|:---:|:---:|
+| **3.7.1** | **Verify that** initiating a fine-tuning run requires authenticated authorization from a designated approver separate from the person requesting the run, with the approval identity, timestamp, and approved training configuration recorded in an immutable audit log before pipeline execution begins. | 2 | D/V |
```
ottosulin: Very good work here, but there is a bit of overlap with existing requirements. 3.4.2 already requires "peer review approval before use in training" (L1). 3.2.3 already requires "immutable audit records" for model changes (L2).

The novel part is specifically requiring a separate designated approver for initiating a fine-tuning run (not just approving artifacts). But this requirement also packs three distinct concerns into one: (1) dual-person authorization, (2) separation of duties, and (3) immutable pre-execution audit logging.

A few options:

- Option A: Narrow this to only the genuinely new requirement  "_Verify that initiating a fine-tuning or retraining run requires authorization from a person who did not request the run_", and drop the audit log clause since 3.2.3 already covers it.
- Option B: Expand 3.4.2 to explicitly cover run initiation approval (not just artifact review).

Also worth noting: no major ML platform (SageMaker, Vertex AI, Azure ML) ships this as a built-in feature afaik. Teams must build custom approval gates, e.g. via CI/CD integration. Thinking because of that should Option A + L3 be the right choice...

1.0/en/0x10-C03-Model-Lifecycle-Management.md
```diff
@@ -85,6 +85,19 @@ Hosted and provider-managed models may change behavior without notice. These con
...
+| **3.7.2** | **Verify that** the reward model used in any RLHF fine-tuning stage is treated as a versioned, signed artifact subject to the same integrity verification and validation gates required of policy model artifacts before it is used in a training run. | 2 | D/V |
```
ottosulin: This is very good addition. Reward models are a distinct artifact class that existing requirements don't address, and in practice they can be often treated as static infrastructure rather than versioned ML artifacts.

That said, 3.1.2 already says "all model artifacts (weights, configurations, tokenizers, base models, fine-tunes, adapters, and safety/policy models) are cryptographically signed." One could argue reward models fall under "all model artifacts" already. The value here is making the implicit explicit, since reward models are consumed during training, not at deployment admission, so 3.1.2's "verified at deployment admission" clause would not catch an unsigned reward model used in a training run.

Suggestion: I would support this requirement, but consider simplifying the language. The phrase "_subject to the same integrity verification and validation gates required of policy model artifacts_" is circular; it depends on whichever other requirements happen to exist. More direct would be something like: "_Verify that reward models used in RLHF fine-tuning are versioned, cryptographically signed, and integrity-verified before use in a training run._"

Implementability is there so L2 appropriate: Sigstore's model-signing library can sign any model artifact, and MLflow supports model versioning.

1.0/en/0x10-C03-Model-Lifecycle-Management.md
```diff
@@ -85,6 +85,19 @@ Hosted and provider-managed models may change behavior without notice. These con
...
+| **3.7.3** | **Verify that** reward hacking is evaluated after each RLHF training stage by measuring the policy model's reward scores on a held-out probe set and comparing them to human preference scores on the same set, with the run blocked from promotion if divergence exceeds a defined threshold. | 3 | D/V |
```
ottosulin: This addresses a real threat, but the requirement is a bit too prescriptive. It mandates a specific detection technique (held-out probe set + human preference comparison + divergence threshold) rather than stating the security objective. There are other valid approaches to detecting reward hacking: KL divergence constraints, information-theoretic methods (e.g., InfoRM), ensemble reward models, or policy rollback triggers based on behavioral anomalies.

Suggestion: Reword to state the objective rather than the method: "_Verify that RLHF training stages include automated detection of reward hacking or reward model over-optimization (e.g., held-out human-preference probe sets, divergence thresholds, or KL penalty monitoring), with the run blocked from promotion if detection thresholds are exceeded._"

L3 is correct, no off-the-shelf tool automates this. TRL (Hugging Face) and OpenRLHF provide RLHF training loops but don't include built-in reward hacking gating. This is still largely a research-stage practice.

1.0/en/0x10-C03-Model-Lifecycle-Management.md
```diff
@@ -85,6 +85,19 @@ Hosted and provider-managed models may change behavior without notice. These con
...
+| **3.7.4** | **Verify that** in multi-stage fine-tuning pipelines, each stage's output checkpoint is cryptographically signed and recorded as a distinct artifact in the model registry before the next stage begins, enabling isolation and rollback of any compromised intermediate stage. | 3 | D/V |
```
ottosulin: The gap argument is reasonable: 3.1.2 covers deployment-time signing, and this targets intra-pipeline intermediate artifacts. However, the overlap concern is real:

- 4.2.2 already requires build artifacts to be cryptographically signed with build-origin metadata (L2).
- 3.1.2 covers "all model artifacts" with signing, which could include checkpoints.
- 3.3.3 covers rollback of complete model state (L2).

The genuinely new aspect is the stage-by-stage gate: sign checkpoint N's output before stage N+1 starts. That is a valid supply chain hardening measure, analogous to SLSA's build provenance requirements applied to ML pipelines.

Suggestion: tighten it to focus on the gate behavior: "_Verify that in multi-stage fine-tuning pipelines, each stage's output is integrity-verified before the next stage consumes it, and intermediate checkpoints are registered as distinct artifacts enabling per-stage rollback._" The signing mechanism is already covered by 3.1.2/4.2.2 and doesn't need to be restated.

L3 is appropriate. Sigstore can technically sign checkpoints, but no ML pipeline orchestrator integrates this as a built-in stage gate today.

PR #601, opened 2026-03-28
feat(C4): add 4.4.6 context window secrets isolation

RicoKomenda:
Gap identified via systematic cross-reference of OWASP AISVS against the Open Agent Security Benchmark (OASB-1, https://www.oasb.ai/controls).

OASB control 5.2 (Context Window Isolation, L1) flags the risk of agents retrieving credentials at runtime and placing them in the model context window, system prompt, or tool call parameters, making them accessible to the model and extractable via prompt injection or model output inspection.

Cross-check against existing AISVS controls:
- C4.4.3 covers secrets not embedded in code/config/artifacts (deployment time)
- C8.3.2 covers credentials not persisted in long-term memory stores
- Appendix C AC.3.1 covers written policy for code-generation workflows only

None of the above cover the runtime enforcement gap: an agent that correctly retrieves a secret from a secrets manager but then injects it into the prompt as a string. This control closes that gap with a technically verifiable requirement scoped to the agent runtime layer.

Ref: OASB-1 control 5.2 - https://www.oasb.ai/controls

### DIFF
--- a/1.0/en/0x10-C09-Orchestration-and-Agentic-Action.md
+++ b/1.0/en/0x10-C09-Orchestration-and-Agentic-Action.md
@@ -13,7 +13,7 @@ Bound runtime expansion (recursion, concurrency, cost) and halt safely on runawa
 > **Scope note:** Controls in this section govern internal orchestration runtime budgets: per-task recursion depth, concurrency, wall-clock time, token spend, and monetary limits. They apply inside the agentic execution layer, not at the API ingress edge (C2.6) and not in response to adversarial probing patterns (C11.4, C11.5). A single token-spend cap at the API gateway does not satisfy C9.1 budget enforcement, which requires enforcement within the orchestration runtime itself.
 
 | # | Description | Level |
-| :--: | --- | :---:|
+| :--: | --- | :---: |
 | **9.1.1** | **Verify that** per-execution budgets (max recursion depth, max fan-out/concurrency, wall-clock time, tokens, and monetary spend) are configured and enforced by the orchestration runtime. | 1 |
 | **9.1.2** | **Verify that** cumulative resource/spend counters are tracked per request chain and hard-stop the chain when thresholds are exceeded. | 2 |
 | **9.1.3** | **Verify that** circuit breakers terminate execution on budget violations. | 2 |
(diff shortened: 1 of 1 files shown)


1.0/en/0x10-C09-Orchestration-and-Agentic-Action.md
```diff
@@ -83,11 +83,12 @@ Protect agent-to-agent and agent-to-tool communications from hijacking, injectio
...
 | **9.6.2** | **Verify that** when an agent acts on a user's behalf, the runtime propagates an integrity-protected delegation context (user ID, tenant, session, scopes) and enforces that context at every downstream call without using the user's credentials. | 2 |
 | **9.6.3** | **Verify that** authorization is re-evaluated on every call (continuous authorization) using current context (user, tenant, environment, data classification, time, risk). | 2 |
 | **9.6.4** | **Verify that** all access control decisions are enforced by application logic or a policy engine, never by the AI model itself, and that model-generated output (e.g., "the user is allowed to do this") cannot override or bypass access control checks. | 2 |
+| **9.6.5** | **Verify that** secrets and credentials required by an agent at runtime are not included in the model context window, system prompt, or tool call parameters, and are instead handled through out-of-band mechanisms (e.g., environment variables, credential proxies, or wrapper scripts) that keep secrets outside the model's observable context. | 2 |
```
jmanico: Typically, environment variables are a bad place for secrets. May I suggest:

_Verify that secrets and credentials required by an agent at runtime are not exposed within the model’s observable context, including the context window, system prompts, or tool call parameters, and are instead provided via out-of-band mechanisms such as **credential proxies, secrets manager injection, runtime sidecar authentication, or short-lived scoped tokens.**_

RicoKomenda: Done :)

vishaljindal1990:
This looks like a solid addition and I agree with moving it to C9.6 — it captures a real runtime gap that’s easy to miss. One suggestion: could we tighten the verification language so it’s clearer what evidence satisfies this control? For example, should compliance be demonstrated via architecture (e.g., sidecars/proxies), inspection of prompts/tool payloads to ensure secrets never enter model-visible context, or test cases that validate this behavior? Making that explicit would help reviewers apply this consistently across implementations.

Issue #588, opened 2026-03-24
Discussion: split compound AISVS requirements that bundle multiple independently testable controls


jmanico:
A review of the current AISVS 1.0 English standard found a recurring standards-design issue: some requirements bundle two or three independently testable controls into a single requirement.

This makes the standard harder to implement, verify, and discuss because teams can comply with one part of a requirement while failing another, yet the requirement still reads like a single atomic control.

This issue is intended as a discussion thread before any PR is drafted.

What counts as a compound requirement?

Typical pattern:
- one requirement combines a preventive control and a detective or logging control
- one requirement combines a producer obligation and a consumer or runtime validation obligation
- one requirement combines multiple enforcement mechanisms that could pass or fail independently
- one requirement combines a primary control plus a required response action

High-confidence split candidates

1. 3.1.2 in 1.0/en/0x10-C03-Model-Lifecycle-Management.md
- currently combines artifact signing, verification at deployment admission, and verification on load

2. 3.2.1 in 1.0/en/0x10-C03-Model-Lifecycle-Management.md
- currently combines input validation testing, output sanitization testing, and safety evaluation with thresholds

3. 4.3.2 in 1.0/en/0x10-C04-Infrastructure.md
- currently combines isolated network segments, no direct internet access, no shared identity roles or security groups, and no cross-environment connectivity

4. 4.8.2 in 1.0/en/0x10-C04-Infrastructure.md
- currently combines model signing during packaging, runtime validation, and rejection of altered or unverified models

5. 8.1.5 in 1.0/en/0x10-C08-Memory-Embeddings-and-Vector-Database.md
- currently combines retrieval logging and prompt-injection scanning before context incorporation

6. 9.4.3 in 1.0/en/0x10-C09-Orchestration-and-Agentic-Action.md
- currently combines tamper-evident logging, independent verifiability, and a large required audit context schema

7. 10.3.3 in 1.0/en/0x10-C10-MCP-Security.md
- currently combines internal-only SSE use, TLS, authentication, schema validation, payload size limits, rate limiting, and prohibition on public internet exposure

8. 13.2.7 in 1.0/en/0x10-C13-Monitoring-and-Logging.md
- currently combines isolate compromised models, block malicious users, and escalate critical events

9. 13.6.4 in 1.0/en/0x10-C13-Monitoring-and-Logging.md
- currently combines automated alerts and retraining or replacement workflow initiation

Additional strong candidates

[...]

PR #614, opened 2026-03-30
fix(C7-C9): split compound requirements per issue #588

RicoKomenda:
Splits each compound requirement into independently testable controls. New control numbers use unused slots within each section.

C7 (2 splits, 3 new controls):
- 7.4.3 -> 7.4.3 (recursion depth) + 7.4.4 (delegation limits) + 7.4.5 (tool allow-list)
- 7.8.1 -> 7.8.1 (attribution present) + 7.8.3 (attribution from metadata, not model)

C8 (3 splits, 3 new controls):
- 8.1.4 -> 8.1.4 (tag at write time) + 8.1.7 (tag immutability)
- 8.3.2 -> 8.3.2 (data minimization) + 8.3.7 (session discard)
- 8.3.6 -> 8.3.6 (security reset capability) + 8.3.8 (quarantine exclusion from retrieval)

C9 (3 splits, 3 new controls):
- 9.3.2 -> 9.3.2 (quotas enforced and logged) + 9.3.7 (fail closed on breach)
- 9.4.3 -> 9.4.3 (tamper-evident log storage) + 9.4.5 (sufficient audit context fields)
- 9.8.4 -> 9.8.4 (isolated context window) + 9.8.7 (dedicated scoped credentials)

Also fixes pre-existing MD060 separator row formatting in C7 and C8.


ottosulin (changes requested):
Nice work on the splits Rico!

Mostly looking good, few items to discuss before merge:

- 7.4.5 (tool allow-list) - I think this overlaps with 9.6.1. Existing 9.6.1 already covers: "restrict which tools an agent may invoke, which parameter values it may supply..." The proposed 7.4.5 is a simpler version of the same control. This creates a "which one do I comply with?" problem. Recommend removing 7.4.5 and adding a cross-reference to C9.6 instead, or explicitly scoping it to non-agentic use cases only, although not sure if needed and how put this into words...

- 7.4.4 (delegation limits) - the wording "preventing agents from re-delegating beyond their authorized scope" sounds to me to be in C9 territory? C7 is about model behavior/output.

- 7.8.3 ordering - should come after 7.8.2 In the current diff, the table reads 7.8.1, 7.8.3, 7.8.2. Move 7.8.3 below 7.8.2 so numbering is correct.

- Appendix D - missing entries for most new IDs The PR updates Appendix D for 9.3.7 and 9.4.5 but does not add entries for 7.4.4, 7.4.5, 7.8.3, 8.1.7, 8.3.7, 8.3.8, or 9.8.7. Some may be covered by existing broad entries, but please verify each new ID appears at least once in the inventory.

- Also consider splitting the Appendix D entry at line 405 ("Tamper-evident audit logs with full reconstruction context") into two entries now that 9.4.3 and 9.4.5 are separate controls - that was the whole point of the split, right?

Issue #609, opened 2026-03-30
Define enforcement boundaries for system instruction integrity controls


vishaljindal1990:
**Problem**
Current control language around “protecting system instructions” is directionally correct but leaves ambiguity in how compliance should be verified across implementations.

**Why this matters**
Without clear verification guidance, different teams may interpret “protection” inconsistently, leading to gaps where system instructions are technically present but still influenceable (e.g., via prompt composition, tool calls, or downstream orchestration).

**Proposed improvement areas**
* Define what constitutes “model-observable context” for system instructions
* Clarify acceptable enforcement mechanisms, such as:
  * Isolation between system and user inputs
  * Runtime detection of override/injection patterns
  * Controls at orchestration/tooling layers
* Provide examples of verifiable evidence (architecture, tests, runtime inspection)

**Goal**
Make the control more testable, implementable, and consistent for assessors.

Related to PR: #607

PR #621, opened 2026-03-30
Clarify verification and enforcement boundaries for instruction hierarchy (C2.1.2)

vishaljindal1990:
This PR builds on #609 and introduces small clarifications to improve verification consistency for instruction hierarchy enforcement.

The current control (C2.1.2) clearly defines expected behavior (system/developer instructions override user input), but verification expectations can be implicit in systems with orchestration layers and multi-step tool interactions.
This change does not introduce new requirements, but:
* Clarifies that enforcement must hold across multi-step and tool-augmented workflows
* Adds explicit language to prevent influence via prompt composition or intermediate outputs
* Adds a small verification note to C2.1.4 to reinforce preservation of system instructions under adversarial conditions

These updates aim to improve implementability and assessor consistency while staying aligned with existing controls in C2 and C9.


RicoKomenda (commented):
Thanks for this. The clarifications to 2.1.2 are directionally right, but a couple of things to address before merge:

**Uppercase MUST**

`This enforcement MUST be preserved` uses RFC 2119 uppercase, which we don't use anywhere else in the standard. The rest of the sentence also drops back to lowercase `must`, so there's an inconsistency within the same control. Please lowercase both to match the rest of the document.

**Separator row in C2.1**

The separator row in the C2.1 table has been expanded to a very long string of dashes (300+). This is likely an IDE formatter artifact. Please revert it to match the existing style used in the rest of the file.

**Unrelated change**

The `.gitignore` addition of `.idea/` is an IDE artifact. Please remove it from this PR or submit it separately.

PR #625, opened 2026-03-31
feat(C5): add just-in-time access provisioning control for AI resources (5.2.9)

RicoKomenda:
## Summary

Adds **5.2.9** to C5.2 (Authorization & Policy) to address a gap identified during OASB cross-referencing and review of NIST SP 800-207 (Zero Trust) and NIST AI RMF guidance.

**New control:**

> **Verify that** privileged access to model weights, training pipelines, and production AI configuration is provisioned on a just-in-time basis with a defined maximum session duration and automatic expiry, every access session is logged with actor identity, justification, and resource scope, and permanent standing privileged access to these resources is not permitted.

**Level:** 2

## Why this is needed

Existing controls (5.2.1, 5.2.2) cover RBAC/ABAC policy and logging of access modifications, but do not require JIT provisioning or prohibit permanent standing access. Permanent standing access to model weights and training pipelines is a significant attack surface: a compromised account with standing access can silently exfiltrate weights, inject poisoned training data, or alter production configuration without any time-bounded detection window.

JIT access is well-supported by PAM tooling (HashiCorp Vault, CyberArk, AWS IAM Identity Center temporary roles, Azure PIM) and is verifiable through access log inspection. This is AI-specific because model weights and training pipelines are the critical assets in an AI system that do not have a direct equivalent in a general web application.

## Changes

- `1.0/en/0x10-C05-Access-Control-and-Identity.md`: add 5.2.9, fix MD060 separator rows


1.0/en/0x10-C05-Access-Control-and-Identity.md
```diff
@@ -32,6 +32,7 @@ Implement access controls for all AI resources with explicit permission models a
 | **5.2.6** | **Verify that** authorization decisions are externalized to a dedicated policy decision point (e.g., OPA, Cedar, or equivalent). | 3 |
 | **5.2.7** | **Verify that** policies evaluate dynamic attributes at runtime including user role or group, resource classification, request context, tenant isolation, and temporal constraints. | 3 |
 | **5.2.8** | **Verify that** policy cache TTL values are defined based on resource sensitivity, with shorter TTLs for high-sensitivity resources, and that cache invalidation capabilities are available. | 3 |
+| **5.2.9** | **Verify that** privileged access to model weights, training pipelines, and production AI configuration is provisioned on a just-in-time basis with a defined maximum session duration and automatic expiry, every access session is logged with actor identity, justification, and resource scope, and permanent standing privileged access to these resources is not permitted. | 2 |
```
ottosulin: Definitely would want higher maturity organizations apply a lot of controls to model weight access. 

I am a little bit tempted to ask this to be pushed to L3 to not pack too much to L2, taking into account everything else. But we can also revisit this discussion a bit later.

One concern: this is a compound requirement that bundles three independently verifiable things: (1) JIT provisioning with session duration/auto-expiry, (2) session logging with justification, and (3) prohibition of standing access.

The JIT + no-standing-access parts are tightly coupled, so keeping them together makes sense. But the logging clause is independently testable and partially overlaps with existing 5.2.2 (which already requires logging of access control modifications with timestamps, actor identities, resource identifiers, and permission changes). 

Another new element is requiring a justification field. Many current products do contain JIT + justification logging in the same feature (e.g. Microsoft products), but I'm not a big fan of including requirements that today require acquiring a commercial product with no good open source options available (Vault open source has JIT but no justification logging). Rolling out your own not really a a realistic option for most orgs. Also. I'm not sure if the justification requirement is exactly a technical requirement, little bit 50/50. So I'm a bit hesitant of the justification part because of these reasons.

Few alternative options I can think of:

Option A: Drop the logging and justification clause from 5.2.9 since 5.2.2 largely covers audit logging already.

Option B: same but keep justification.

Option C: add justification logging as separate L3 requirement.

PR #626, opened 2026-03-31
feat(C6): add publisher key pinning control for trusted source enforcement (6.4.6)

RicoKomenda:
## Summary

Adds **6.4.6** to C6.4 (Trusted Source Enforcement) to close a gap where 6.4.2 requires signature validation but does not require pinning the publisher's signing key per source registry.

**New control:**

> **Verify that** cryptographic signing keys used to authenticate model publishers are pinned per source (e.g., Hugging Face, internal registry), that key rotation events require explicit re-approval before updated keys are trusted, and that artifacts signed by unknown or revoked keys are rejected.

**Level:** 2

## Why this is needed

6.4.2 says to validate signatures, but without key pinning an attacker who compromises a registry or performs a DNS hijack can substitute their own signing key. Key pinning closes this: organizations maintain an explicit allowlist of trusted publisher keys per source registry, and any key not on that list causes the artifact to be rejected. Key rotation is treated as a security event requiring explicit re-approval rather than being transparently accepted.

This is analogous to certificate pinning in TLS but applied to the AI supply chain (model weights, containers). It is AI-specific because model weight repositories (Hugging Face, model hubs) are not covered by standard web PKI pinning mechanisms.

## Changes

- `1.0/en/0x10-C06-Supply-Chain.md`: add 6.4.6, fix MD060 separator rows
- `1.0/en/0x93-Appendix-D_AI_Security_Controls_Inventory.md`: add entry to AD.6


1.0/en/0x10-C06-Supply-Chain.md
```diff
@@ -54,12 +54,13 @@ Pin every dependency to immutable digests and verify builds to guarantee tamper-
...
+| :--------: | ------------------------------------------------------------------------------------------------------------------- | :---: |
 | **6.4.1** | **Verify that** model weights, datasets, and containers are downloaded only from approved sources or internal registries. | 1 |
 | **6.4.2** | **Verify that** cryptographic signatures validate publisher identity before artifacts are cached locally. | 1 |
 | **6.4.3** | **Verify that** egress controls block unauthenticated artifact downloads to enforce trusted-source policy. | 2 |
 | **6.4.4** | **Verify that** repository allow-lists are reviewed periodically with evidence of business justification for each entry. | 3 |
 | **6.4.5** | **Verify that** policy violations trigger quarantining of artifacts and rollback of dependent pipeline runs. | 3 |
+| **6.4.6** | **Verify that** cryptographic signing keys used to authenticate model publishers are pinned per source (e.g., Hugging Face, internal registry), that key rotation events require explicit re-approval before updated keys are trusted, and that artifacts signed by unknown or revoked keys are rejected. | 2 |
```
ottosulin: The proposed level is 2, but I think this should be Level 3 for two reasons:

Tooling maturity: I looked into the current state. Sigstore/model-signing v1.0 (April 2025) and the OpenSSF Model Signing (OMS) spec (June 2025) support identity-based verification (`--identity` + `--identity_provider`), but neither provides a built-in "pin this key for this source registry" feature. Organizations would need to build a custom policy layer on top: maintaining per-registry key allowlists, wiring up rotation re-approval workflows, and enforcing rejection of unknown keys. That is significant custom engineering.

Level consistency within C6.4: The neighboring controls at L3 are comparable in difficulty:

- 6.4.4 (periodic allow-list review with business justification) = L3
- 6.4.5 (quarantine + rollback on policy violations) = L3
Key pinning with rotation re-approval is at least as complex as those. Placing it at L2 while those are L3 creates an inconsistency.


Also, the current text packs three things into one requirement:

- Key pinning per source registry
- Key rotation re-approval workflow
- Rejection of artifacts signed by unknown or revoked keys
Item (3) is arguably already covered by 6.4.2 ("cryptographic signatures validate publisher identity"). If you validate against a known identity and the key is unknown or revoked, the validation should fail. Including it here creates overlap.

I would suggest trimming to focus on what is new taking the above into account:

> Verify that cryptographic signing keys used to authenticate model publishers are pinned per source registry (e.g., Hugging Face, internal registry) and that key rotation events require explicit re-approval before updated keys are trusted.

This keeps the requirement focused on the two novel aspects (pinning + rotation governance) and avoids restating what 6.4.2 already requires.

PR #629, opened 2026-03-31
feat(C9): add A2A task handoff identity validation control (9.5.5)

RicoKomenda:
## Summary

Adds **9.5.5** to C9.5 (Secure Messaging and Protocol Hardening) to address trust boundary enforcement during structured agent-to-agent task delegation.

**New control:**

> **Verify that** when agents accept structured task handoffs from peer agents (e.g., via A2A or equivalent delegation protocol), the receiving agent validates the sending agent's identity and authorized scope against an approved agent registry before accepting the task, and rejects handoffs from agents not present in or explicitly excluded from that registry.

**Level:** 2

## Why this is needed

Structured agent-to-agent delegation protocols (Google A2A, OpenAI's emerging multi-agent patterns, AutoGen, LangGraph) allow one agent to hand off tasks with full context to another. This creates a lateral movement path: a compromised or malicious agent can impersonate a legitimate orchestrator and inject tasks into a trusted downstream agent's execution context.

Existing C9.5 controls cover channel encryption and mutual authentication (9.5.1), schema validation (9.5.2), and replay protection (9.5.3), but none require the receiving agent to validate the sender against an explicit approved registry. An agent that accepts tasks from any authenticated peer is vulnerable to task injection from a compromised peer that passed authentication but was never supposed to delegate to it. The approved registry creates a second check: even a valid identity must be on the pre-approved list of agents authorized to delegate.

This is AI-specific because task delegation in multi-agent systems has no direct analogue in traditional web service authorization: the "message" is an agent task with a goal and partial context, not a simple API call, and the downstream agent may take significant actions based on it.

## Changes

- `1.0/en/0x10-C09-Orchestration-and-Agentic-Action.md`: add 9.5.5


1.0/en/0x10-C09-Orchestration-and-Agentic-Action.md
```diff
@@ -75,6 +75,7 @@ Protect agent-to-agent and agent-to-tool communications from hijacking, injectio
 | **9.5.2** | **Verify that** all messages are strictly schema-validated; unknown fields, malformed payloads, and oversized frames are rejected. | 1 |
 | **9.5.3** | **Verify that** message integrity covers the full payload including tool parameters, and that replay protections (nonces/sequence numbers/timestamp windows) are enforced. | 2 |
 | **9.5.4** | **Verify that** agent outputs propagated to downstream agents are validated against semantic constraints (e.g., value ranges, logical consistency) in addition to schema validation. | 2 |
+| **9.5.5** | **Verify that** when agents accept structured task handoffs from peer agents (e.g., via A2A or equivalent delegation protocol), the receiving agent validates the sending agent's identity and authorized scope against an approved agent registry before accepting the task, and rejects handoffs from agents not present in or explicitly excluded from that registry. | 2 |
```
ottosulin: The threat you describe (lateral task injection from a compromised but authenticated peer) is worth addressing.

However, I see a few issues:

- Overlap: Parts (a) and (b) of this requirement -validating sender identity and authorized scope - are already covered by 9.4.1 + 9.5.1 (identity/mutual auth) and 9.6.1 + 9.6.3 (fine-grained authorization/continuous re-evaluation). The new concept here is the "approved agent registry" as a second check beyond authentication.

- Wrong section: This is an authorization control, not a messaging/protocol control. I think this belongs in C9.6 (Authorization, Delegation, and Continuous Enforcement), not C9.5 (Secure Messaging and Protocol Hardening). C9.5 covers channel-level concerns (TLS, schema, replay). An agent allowlist is about who may delegate to whom, which is authorization.

- Compound requirement: The current text bundles identity validation + scope validation + registry check into one requirement. Per our guidelines, each requirement should test one thing. Since identity and scope are already covered, I'd suggest scoping this down to just the net-new registry/allowlist concept.

So building on your proposal with these in mind, suggestion for 9.6.6 in C9.6:

> Verify that agent-to-agent task delegation is restricted by an explicit peer authorization policy (e.g., an approved agent registry or allowlist) so that even authenticated agents can only delegate to or accept delegations from pre-approved peers, with delegation attempts from unlisted agents rejected by default.

L2 seems still right for this. Thoughts?

PR #630, opened 2026-03-31
feat(C10): add MCP tool namespace collision prevention control (10.6.5)

RicoKomenda:
## Summary

Adds **10.6.5** to C10.6 (Transport Restrictions & High-Risk Boundary Controls) to address a gap where no control prevents a malicious or compromised MCP server from shadowing trusted tool definitions in a multi-server setup.

**New control:**

> **Verify that** when multiple MCP servers are connected simultaneously, the client or orchestration layer maintains an authoritative tool namespace mapping that detects tool name collisions and prevents a lower-trust or newly registered server from shadowing or overriding tool definitions from a higher-trust server.

**Level:** 2

## Why this is needed

When an MCP client connects to multiple servers, all servers contribute to a shared tool namespace via `tools/list`. If two servers expose a tool with the same name, the client must decide which definition takes precedence. An attacker who can register a new MCP server (or compromise an existing one) can shadow a high-trust tool (e.g., `code_execute`, `file_write`, `send_email`) with a malicious version that the LLM invokes instead of the legitimate one.

This attack was documented in security research on "tool poisoning" in multi-MCP environments published in early 2025 and is discussed in the OWASP MCP Security Cheat Sheet.

Existing controls (10.4.5) track tool definition hash changes but only within a single server. 10.6.2 prohibits dynamic dispatch but does not address cross-server collision. 10.6.5 is the missing piece: an authoritative namespace map with trust-ranked collision resolution.

## Changes

- `1.0/en/0x10-C10-MCP-Security.md`: add 10.6.5


1.0/en/0x10-C10-MCP-Security.md
```diff
@@ -85,6 +85,7 @@ Ensure secure discovery, authentication, authorization, transport, and use of MC
 | **10.6.2** | **Verify that** MCP servers expose only allow-listed functions and resources and prohibit dynamic dispatch, reflective invocation, or execution of function names influenced by user or model-provided input. | 3 |
 | **10.6.3** | **Verify that** tenant boundaries, environment boundaries (e.g., dev/test/prod), and data domain boundaries are enforced at the MCP layer to prevent cross-tenant or cross-environment server or resource discovery. | 3 |
 | **10.6.4** | **Verify that** MCP security controls enforce fail-closed semantics: if a signature verification, authentication check, or policy evaluation fails or cannot be completed, the default action is to deny the request rather than permit it. | 2 |
+| **10.6.5** | **Verify that** when multiple MCP servers are connected simultaneously, the client or orchestration layer maintains an authoritative tool namespace mapping that detects tool name collisions and prevents a lower-trust or newly registered server from shadowing or overriding tool definitions from a higher-trust server. | 2 |
```
jmanico: May I suggest one of these?

_Verify that the client maintains an authoritative tool namespace mapping that detects collisions and prevents lower-trust servers from shadowing higher-trust tool definitions._

or

_Verify that the client or orchestration layer maintains an authoritative tool namespace mapping that detects tool name collisions and prevents a lower-trust or newly registered server from shadowing tool definitions from a higher-trust server._

ottosulin: Also, the requirement asks the client or orchestration layer to maintain an "authoritative tool namespace mapping" with trust-ranked collision resolution. As of today:

- The MCP specification does not mandate tool namespacing (PR 334 for optional @-prefixed namespaces has stalled).
- No mainstream MCP client (Claude Desktop, Cursor, Claude Code) implements trust-ranked collision resolution out of the box. Clients vary between "last registered wins," alphabetical, or outright failure on duplicates.
- No standardized framework or library exists for authoritative namespace mapping.

This control currently requires custom implementation with no off-the-shelf support, which is more consistent with Level 3 ("makes sense conceptually but lacks mature tooling or framework support"). Compare with 10.6.1 (stdio restrictions, Level 3) and 10.6.2 (no dynamic dispatch, Level 3) in the same section - those are arguably simpler to implement than trust-ranked namespace governance, yet they are already Level 3.

Recommendation: change to Level 3 for consistency with the section and the current state of tooling. If/when the MCP spec adopts mandatory namespacing, we can promote it to Level 2.

jmanico: .... this is something folks who care about security need to implement. I see this in some security guides. I'm ok with level 3, but I'd also be ok with level 2.

jmanico (approved):

PR #634, opened 2026-03-31
feat(C3): add parallel deployment cohort isolation control (3.3.5)

RicoKomenda:
## Summary

Adds **3.3.5** to C3.3 (Controlled Deployment & Rollback) to address a gap where no control prevents cross-contamination between model versions running simultaneously in A/B, canary, or shadow deployments.

**New control:**

> **Verify that** model versions running in parallel (e.g., A/B tests, canary deployments, shadow deployments) are isolated such that prompt caches, session state, and retrieval context are not shared across cohorts, preventing cross-version data contamination and behavioral inference across deployment groups.

**Level:** 2

## Why this is needed

Canary and A/B deployments route requests to different model versions simultaneously. If shared infrastructure (KV caches, session stores, vector retrieval indices) leaks between cohorts, a user assigned to the stable model can receive responses influenced by the canary model's cached state, and vice versa. This creates both a security boundary violation (canary model state visible to stable model users) and a data integrity problem (behavioral measurements are contaminated across groups).

Existing C3.3 controls cover rollout mechanics (3.3.1) and rollback (3.3.2) but say nothing about isolation between concurrently running versions. This is AI-specific because prompt caches, embedding caches, and retrieval indices are AI-layer constructs with no direct equivalent in standard blue/green web deployments.

## Changes

- `1.0/en/0x10-C03-Model-Lifecycle-Management.md`: add 3.3.5, fix MD060 separator rows
- `1.0/en/0x93-Appendix-D_AI_Security_Controls_Inventory.md`: add entry to AD.13

### DIFF
--- a/1.0/en/0x10-C03-Model-Lifecycle-Management.md
+++ b/1.0/en/0x10-C03-Model-Lifecycle-Management.md
@@ -48,7 +48,7 @@ Model deployments must be controlled, monitored, and reversible.
 | **3.3.2** | **Verify that** rollback capabilities restore the complete model state (weights, configurations, dependencies including adapters and safety/policy models) atomically. | 2 |
 | **3.3.3** | **Verify that** emergency model shutdown capabilities can disable model endpoints within a pre-defined response time. | 3 |
 | **3.3.4** | **Verify that** emergency shutdown cascades to all parts of the system including e.g. deactivating agent tool and MCP access, RAG connectors, database and API credentials, and memory-store bindings. | 3 |
-| **3.3.5** | **Verify that** model versions running in parallel (e.g., A/B tests, canary deployments, shadow deployments) are isolated such that prompt caches, session state, and retrieval context are not shared across cohorts, preventing cross-version data contamination and behavioral inference across deployment groups. | 2 |
+| **3.3.5** | **Verify that** model versions running in parallel (e.g., A/B tests, canary, shadow deployments) use isolated runtime state so that AI-specific shared resources (e.g., KV caches, prompt caches, session state, retrieval indices) are not shared across deployment cohorts. | 2 |
 
 ---
 


1.0/en/0x10-C03-Model-Lifecycle-Management.md
```diff
@@ -43,11 +43,12 @@ Models must pass defined security and safety validations before deployment.
...
 | **3.3.1** | **Verify that** production deployments implement gradual rollout mechanisms (e.g., canary or blue-green deployments) with automated rollback triggers based on pre-agreed error rates, latency thresholds, guardrail alerts, or tool/MCP failure rates. | 2 |
 | **3.3.2** | **Verify that** rollback capabilities restore the complete model state (weights, configurations, dependencies including adapters and safety/policy models) atomically. | 2 |
 | **3.3.3** | **Verify that** emergency model shutdown capabilities can disable model endpoints within a pre-defined response time. | 3 |
 | **3.3.4** | **Verify that** emergency shutdown cascades to all parts of the system including e.g. deactivating agent tool and MCP access, RAG connectors, database and API credentials, and memory-store bindings. | 3 |
+| **3.3.5** | **Verify that** model versions running in parallel (e.g., A/B tests, canary deployments, shadow deployments) are isolated such that prompt caches, session state, and retrieval context are not shared across cohorts, preventing cross-version data contamination and behavioral inference across deployment groups. | 2 |
```
ottosulin: Very nice addition. One thing that I'm thinking: this reads as a compound requirement covering three different isolation concerns: (1) prompt/KV caches, (2) session state, and (3) retrieval context. These can fail independently: an org might properly partition sessions but share a vector index across cohorts.

I would suggest simplifying the requirement to focus on the core isolation principle and move the examples into parenthetical guidance, also making this more future-proof. Something like:

> Verify that model versions running in parallel (e.g., A/B tests, canary, shadow deployments) use isolated runtime state so that AI-specific shared resources (e.g., KV caches, prompt caches, session state, retrieval indices) are not shared across deployment cohorts.

This keeps it as one testable assertion ("is runtime state isolated?") with examples rather than three separate things to verify. It also drops "preventing cross-version data contamination and behavioral inference across deployment groups" which is rationale, not a testable condition. The "verify that" should describe the control, not the threat it mitigates.

I am a bit 50/50 should this be bumped to L3 because this requires more than a bit of engineering to really get there:
- Prompt cache isolation: Most serving frameworks (vLLM, TGI, Triton) do not have built-in "cohort-aware" cache partitioning out of the box. You would need separate serving instances per cohort, which is doable but adds operational complexity.
- Session state isolation: Achievable by partitioning session stores by deployment cohort, standard engineering practice.
- Retrieval context isolation: If using shared vector DBs for RAG, you would need namespace separation per model version, which is supported by most vector DBs (Pinecone namespaces, Weaviate tenants, etc.) but requires deliberate architecture.

On the other hand, L2 feels more in line with other L2 requirements + doable without custom components (if not easy) so I think L2 can also be justified.

jmanico (approved):

PR #638, opened 2026-04-03
refactor(C10): reduce MCP Security requirements from 41 to 30 (#636)

### LINKED ISSUE #636: Requirement Reduction Pass
We now have well over 400 requirements and that's a bit out of control. Let's keep going. But we need to do a live final pass and heavily reduce the requirement count. By over 100. I am open to other opinions! :)

cy10101:
## Summary

Reduces C10 (MCP Security) from 41 to 30 requirements (-27%) as part of the requirement reduction pass (#636).

## Approach

Two categories of reduction:

### Within-C10 merges (5 requirements removed)
- **10.2.1 + 10.2.2 + 10.2.5 → 10.2.1**: OAuth client presentation, server validation, and resource-server-only behavior are one implementation requirement, not three.
- **10.3.1 + 10.3.2 → 10.3.1**: "Use encrypted streamable-HTTP" and "use TLS 1.3" are the same control at different abstraction levels.
- **10.3.3 + 10.3.6 → 10.3.3**: Both SSE hardening requirements merged into one.
- **10.4.9 + 10.4.10 + 10.4.11 → 10.4.11**: Per-message signing, replay protection (nonce/timestamp), and server response signing consolidated. TLS already provides channel integrity; the remaining requirement covers response-level signing with freshness for defense-in-depth.

### Cross-chapter deduplication with C9 (6 requirements removed)
- **10.2.4** removed → duplicate of C9.6.4 (model output cannot override access control)
- **10.5.2** removed → duplicate of C9.1.1/C9.1.3 (execution budgets + circuit breakers)
- **10.5.3** removed → duplicate of C9.2.1 (human approval gate for high-risk actions)
- **10.6.1** removed → duplicate of C9.3.1 (sandbox isolation) + already restricted by 10.3.1
- **10.6.3** removed → duplicate of C9.8.1 (tenant/environment isolation)

Added scope notes to C10.2, C10.5, and C10.6 referencing the applicable C9 controls so readers can trace the coverage.

## What this does NOT change
- No new requirements added
- No level changes
- No renumbering (kept original IDs for diff clarity)
- C10.1 and C10.4 core validation requirements untouched

Fixes #636 (partial — C10 chapter)


jmanico:
Hey @RicoKomenda can you eyeball this if you have time? I want to merge this. Any thoughts on this?

RicoKomenda:
Hey @jmanico , looked through it. The overall approach is solid and @cy10101 has done careful work here. A few thoughts before merging:

**The cross-chapter deduplication is mostly right.** 10.5.2, 10.5.3, and 10.6.3 are genuine duplicates of C9 controls and removing them makes sense. The one I would flag is 10.6.1 (stdio local-only enforcement with terminal injection prevention) - C9.3.1 covers plugin sandboxing generally, but the stdio terminal injection vector is specific to MCP's local transport and not really addressed by the C9 control. Worth keeping or at least discussing before removing.

**The within-C10 merges are reasonable** except that 10.3.1+10.3.2 and 10.3.3 touch the same section as our open PR #631 (MCP minimum version enforcement, adding a new 10.3.6). That PR will need a rebase if this lands first, but there is no semantic conflict.

**On timing:** Otto raised on #636 that the raw count is less of a concern than scope creep, and that most requirements are simply not applicable to orgs that don't train their own models. We are discussing #636 tomorrow on the leader call. My suggestion would be to hold this until after that call so we align on the reduction strategy across all chapters rather than doing C10 in isolation. That also gives us a chance to land #630 and #631 cleanly before the reduction pass reshuffles things.

Short version: the PR is good work, but let's sync tomorrow and merge it as part of a coordinated pass.

jmanico:
Good instinct @RicoKomenda I'm with you.

razashariff:
Thanks for taking this on @cy10101 — the C9 cross-chapter dedup and the TLS 1.3 / OAuth groupings all look sensible.

Quick thought on the **10.4.9 + 10.4.10 + 10.4.11** merge though — would it be worth keeping those three separate? They look similar on the surface but they're actually covering different directions and threats:

- 10.4.9 = request signing (client → server)
- 10.4.10 = replay/freshness (nonce + window — orthogonal to signing, a valid signed message is still valid if replayed)
- 10.4.11 = response signing (server → client)

TLS only helps hop-by-hop, so in MCP deployments with gateways or policy sidecars the body is exposed at each hop. That's the gap these three fill between them.

Happy to help reword if it's the verbosity that's the issue rather than the coverage — maybe tightening each to one line would hit the reduction goal without losing them?

Cheers,
Raza

ottosulin:
Thanks @cy10101 for the good PR! And apologies for hijacking this now a bit for further changes :) decided to rather commit to this branch than start a new PR later for the chapter.

You can see the changes in that one commit I made, summarizing them below. TL;DR: don't repeat ASVS, unless it is specifically MCP protocol implementation concern.

Changes in 75e340c :

Further reduces C10 from 30 to 26 requirements and adjusts levels/wording based on ASVS v5 overlap and comparison to current state of MCP spec and implementations.

Removals (4 requirements):

- 10.2.10 (OAuth proxy consent) - generic OAuth consent control, covered by ASVS v5 V10.7.1/V10.4.7
- 10.2.14 (reject broad scopes) - generic OAuth scope minimization, covered by ASVS v5 V10.4.11/V10.2.3
- 10.4.7 (reject duplicate JSON keys) - generic JSON parser security, not MCP-specific
- 10.4.6 old wording replaced (see rewording below)

Level changes (2 requirements):

- 10.3.1 L2 to L1 - transport selection (streamable-HTTP in production, restrict stdio/SSE) is a foundational MCP deployment decision; also trimmed "TLS 1.3 or later with certificate validation" since that's covered by ASVS v5 V12
- 10.4.2 L2 to L3 - schema signing has no current MCP spec or tooling support
- 10.4.11 L2 to L3 - per-message response signing with nonce/replay protection is not in the MCP spec (open proposal PR #2395)

Wording changes:

- C10.2 scope note - added ASVS v5 V10 cross-reference for OAuth details
- 10.3.1 - trimmed TLS version specifics (defer to ASVS for all TLS details)
- 10.4.6 - reworded to emphasize the AI-specific risk: information leakage exploitable via the model context
- 10.6.4 - replaced generic "signature verification" with MCP-specific failure modes (tool schema validation, MCP-Protocol-Version negotiation)

RicoKomenda (approved):

Issue #640, opened 2026-04-03
Add concurrent execution safety requirements to C9.8 (TOCTOU race conditions between agents)


Joe-B-Security:
C9.8 covers runtime isolation, emergent behaviour detection, memory and context window isolation, rate limits, and shutdown capabilities. It does not address race conditions between concurrently executing agents that operate on shared mutable state e.g. when multiple agents hold overlapping permissions on the same state and authorisation enforcement depends on timing between them.

C9.6.3 requires authorisation to be re-evaluated on every call, but it does not address what happens when two agents authorisation checks and state mutations interleave concurrently. C9.8.2 detects emergent behaviour like deadlocks and oscillation, but not TOCTOU exploitation across agents. The C9.2 wiki discusses TOCTOU in the context of approval binding, where the race is between approval and execution within a single workflow and the fix is cryptographic parameter binding. The pattern described here is somewhat different, multiple agents writing to shared state concurrently, where the race is between one agent's grant and another agent's revoke, and the timing window between them can be exploited by an attacker or cause an inadvertent security incident through normal use, the actors, the attack, and the mitigation are all distinct from the approval-binding case.

More discussion here: https://labs.snyk.io/resources/threat-modeling-best-defense/#threat-modeling-black-swan-security-events

I propose something along the lines of:

| # | Description | Level |
|---|---|:---:|
| **9.8.7** | Verify that when multiple agents operate concurrently on shared mutable state, authorization checks and the state mutations they gate are executed atomically (e.g., via serialization, optimistic locking, or compare-and-swap), preventing time-of-check-time-of-use (TOCTOU) vulnerabilities where one agent's state change invalidates another agent's prior authorization check before its action completes. | 2 |

PR #641, opened 2026-04-04
feat(C9): add concurrent execution TOCTOU safety control (9.8.8) (#640)

Joe-B-Security:
PR related to: https://github.com/OWASP/AISVS/issues/640

(Original issue suggested 9.8.7 but I missed it as was out of order, so I used 9.8.8)


RicoKomenda:
TOCTOU is an established security word. We just didn't have it in the spellcheck list yet. Everythings fine now. :)

Issue #639, opened 2026-04-03
Gap analysis: governance, data classification, and operational controls identified via AISec-Playbook comparison


jmanico:
## Summary

A cross-reference of [AISec-Playbook](https://github.com/ferranverdes/AISec-Playbook) (145 controls, 8 domains) against AISVS 1.0 surfaced several areas where AISVS lacks explicit verifiable requirements despite covering the underlying concepts in wiki research or tooling references.

AISVS has significantly deeper technical coverage — particularly MCP security (C10), adversarial robustness (C11), hardware/TEE security (C4.7-4.8), and agent orchestration (C9). The gaps below are limited to technically testable requirements.

## Data Lifecycle Controls

| Missing Area | Potential Chapter | Notes |
|---|---|---|
| **Data Classification & Handling Rules** | C1 | C1 focuses on integrity/traceability but lacks an explicit requirement to classify AI data assets (PII, PHI, secrets, prompts, logs, retrieval corpora) with per-class handling rules |
| **Data Normalization** | C1 | No requirement to standardize encodings, units, timestamps, formats as a security/integrity control before ingestion |
| **Dataset Deduplication & Balance** | C1 | C1.4.5 covers label skew but no explicit deduplication or representation balance requirement |

## Licensing & IP Tracking

| Missing Area | Potential Chapter | Notes |
|---|---|---|
| **License Tracking for Datasets, Models, and Generated Outputs** | C6 | C6.5 wiki discusses licensing but no explicit requirement to track and enforce license compatibility across datasets, models, code, and generated outputs |

## RAG Source Freshness

| Missing Area | Potential Chapter | Notes |
|---|---|---|
| **Source Freshness & Staleness Control** | C8 | No requirement to define freshness requirements for RAG sources or mark when retrieved evidence may be outdated |

## Granular Safe Degradation

| Missing Area | Potential Chapter | Notes |
|---|---|---|
| **Granular Degradation Modes** | C14 or C3 | C14.1 has a kill-switch but lacks verifiable requirements for granular degradation: disable specific tools, switch to read-only, remove a single retrieval source, fall back to a safer model |

## Reference

- AISec-Playbook: https://github.com/ferranverdes/AISec-Playbook
- Comparison methodology: domain-by-domain mapping of all 145 Playbook controls against AISVS C1-C14 + Appendix C requirements and wiki research

RicoKomenda:
@jmanico 
Based on #649, should we close this issue or do you want to check the other controls again?

PR #650, opened 2026-04-05
feat(C14): add granular degradation policy requirement (14.1.5)

### LINKED ISSUE #649: Add granular degradation policy requirement to C14.1
## Summary

AISVS has a kill-switch (C14.1.1), safe-mode rollback (C14.1.3), and several risk-adaptive responses that allow degraded capabilities (C2.8.3, C11.2.2, C11.6.3, C3.3.4). What is missing is a verifiable requirement for a **pre-defined degradation policy**: operators should declare in advance what the system does at each degradation level, rather than relying on ad hoc responses at incident time.

## Proposed control (C14.1.x, L2)

> **Verify that** a documented, pre-defined degradation policy exists that defines at least two intermediate states between full operation and complete shutdown (e.g., disable specific tools or MCP servers, remove a retrieval source, switch to a safer or smaller model, enforce read-only mode), and that each state has defined triggers, scope, and recovery criteria.

## Why this is a genuine gap

- C14.1.1: full kill-switch (binary, on/off)
- C14.1.3: rollback to previous version or safe-mode, but no requirement that safe-mode is pre-defined
- C2.8.3: risk-adaptive responses are allowed but not required to be pre-documented
- C3.3.4: cascading shutdown of all components (also binary)

None of these require that intermediate degradation states be specified, tested, and documented before an incident occurs.

## Design principle check

- AI-specific: yes, the degradation states are AI-layer concerns (model swap, retrieval source removal, tool disabling)
- Security/safety: yes, an undefined degradation path is a safety risk during incidents
- Verifiable: yes, the policy document and its mapping to system behaviour can be audited
- Single concern: yes, one testable assertion (does a pre-defined policy with intermediate states exist?)

Relates to: #639

RicoKomenda:
## Summary

Adds a new control 14.1.5 to C14.1 (Kill-Switch & Override Mechanisms) requiring a documented, pre-defined degradation policy with intermediate states between full operation and complete shutdown.

## Gap addressed

AISVS already has:
- C14.1.1: full kill-switch (binary on/off)
- C14.1.3: rollback to previous version or safe-mode
- C2.8.3: risk-adaptive responses (disable tools, shrink context, HITL)
- C3.3.4: cascading shutdown of all components

What was missing is a requirement for operators to define and document intermediate degradation states *before* an incident, including triggers, scope, and recovery criteria for each state. Without this, teams are left making ad hoc decisions under pressure.

## New control

| # | Description | Level |
|---|---|:---:|
| **14.1.5** | **Verify that** a documented degradation policy defines at least two intermediate states between full operation and complete shutdown (e.g., disabling specific tools or MCP servers, removing a retrieval source, switching to a safer or smaller model, enforcing read-only mode for agents), and that each state has defined triggers, scope, and recovery criteria. | 2 |

## Design principle check

- AI-specific: yes, the degradation states reference AI-layer components (model swap, retrieval source removal, tool disabling, agent mode)
- Security/safety: yes, an undefined degradation path is a safety risk during incidents
- Verifiable: yes, the policy document and its mapping to system behaviour can be audited
- Single concern: yes, one testable assertion (does a pre-defined policy with intermediate states exist?)

Closes #649
Relates to #639

### DIFF
--- a/1.0/en/0x10-C14-Human-Oversight.md
+++ b/1.0/en/0x10-C14-Human-Oversight.md
@@ -16,7 +16,7 @@ Provide shutdown or rollback paths when unsafe behavior of the AI system is obse
 | **14.1.2** | **Verify that** override controls are accessible to only to authorized personnel. | 1 |
 | **14.1.3** | **Verify that** rollback procedures can revert to previous model versions or safe-mode operations. | 3 |
 | **14.1.4** | **Verify that** override mechanisms are tested regularly. | 3 |
-| **14.1.5** | **Verify that** a documented degradation policy defines at least two intermediate states between full operation and complete shutdown (e.g., disabling specific tools or MCP servers, removing a retrieval source, switching to a safer or smaller model, enforcing read-only mode for agents), and that each state has defined triggers, scope, and recovery criteria. | 2 |
+| **14.1.5** | **Verify that** the system can be placed into at least two intermediate operational states between full operation and complete shutdown (e.g., disabling specific tools or MCP servers, removing a retrieval source, switching to a safer or smaller model, enforcing read-only mode for agents), and that each state has defined entry triggers and can be exited independently without requiring a full system restart or shutdown. | 2 |
 
 ---
 
--- a/1.0/en/0x10-C14-Human-Oversight.md
+++ b/1.0/en/0x10-C14-Human-Oversight.md
@@ -16,6 +16,7 @@ Provide shutdown or rollback paths when unsafe behavior of the AI system is obse
 | **14.1.2** | **Verify that** override controls are accessible to only to authorized personnel. | 1 |
 | **14.1.3** | **Verify that** rollback procedures can revert to previous model versions or safe-mode operations. | 3 |
 | **14.1.4** | **Verify that** override mechanisms are tested regularly. | 3 |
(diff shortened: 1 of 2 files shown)


1.0/en/0x10-C14-Human-Oversight.md
```diff
@@ -16,6 +16,7 @@ Provide shutdown or rollback paths when unsafe behavior of the AI system is obse
 | **14.1.2** | **Verify that** override controls are accessible to only to authorized personnel. | 1 |
 | **14.1.3** | **Verify that** rollback procedures can revert to previous model versions or safe-mode operations. | 3 |
 | **14.1.4** | **Verify that** override mechanisms are tested regularly. | 3 |
+| **14.1.5** | **Verify that** a documented degradation policy defines at least two intermediate states between full operation and complete shutdown (e.g., disabling specific tools or MCP servers, removing a retrieval source, switching to a safer or smaller model, enforcing read-only mode for agents), and that each state has defined triggers, scope, and recovery criteria. | 2 |
```
jmanico: This says the verifier needs to verify documentation, which we mostly want to avoid.

Do you want this merged as-is, or would you object to taking the "documented" out of this so its more of a technical and less of a governance requirement?

jmanico (approved):

Issue #102, opened 2025-11-25
Update Contributors and Reviewers


khalidwalidalamri:
This issue is to update Contributors and Reviewers.

razashariff:
Hi Jim — happy to be listed in the AISVS 1.0 contributors when you get to that pass. Contributed C10 via #598 (merged as #608): message-level signing, replay protection, response integrity, autonomous agent identity, and fail-closed semantics.

Thanks for pulling this together for Vienna — looking forward to the 1.0 tag.

Raza Sharif
CyberSecAI Ltd

RicoKomenda:
## Updated Contributors and Reviewers (current as of April 2026)

Refreshed contributor counts from `git shortlog -sn HEAD -- "1.0/en/0x10-C*.md"` to reflect all work to date.

**Project Leads:**
1. Jim Manico
2. Otto Sulin
3. Rico Komenda
4. Russ Memisyazici

**Contributors** (direct requirement authors, ordered by contribution volume):
- Joshua Beck (26 commits)
- b1oo (15 commits)
- Khalid Al-Amri (13 commits, across Khalidwalamri / khalidwalidalamri / Khalid Al-Amri)
- Martin Bjerke (8 commits)
- Almog Langleben (6 commits, across almogbhl / Almog Langleben)
- mattijs moens (5 commits)
- RL Thornton (5 commits)
- Jim Schwoebel (4 commits)
- Joe-B-Security (3 commits)
- Vineeth Sai Narajala (2 commits)
- Tametomo (2 commits)
- Raza Sharif (1 commit — 5 MCP requirements via PR #608, authored under local git identity)
- Vishal Jindal (1 commit)
- William Jawad (1 commit)
- Hari Mukundhan (1 commit)
- Jerry Hoff (1 commit)
- hackwither (1 commit)
- RogueValley (1 commit)
- Uncle Joe (1 commit)
- vatsalgupta (1 commit)

**Reviewers** (CI, spelling, docs, tooling):
- Tetsuo Seto (13 commits)
- deepakrpandey12 / Deepak Pandey (3 commits)
- Starr Brown (1 commit)
- Stuart Small (1 commit)
- Sandhya (1 commit)

deepakrpandey12:
@jmanico can I be added as a Contributor (direct requirement authors, ordered by contribution volume) instead of Reviewer (CI, spelling, docs, tooling) ? 

[...]

PR #661, opened 2026-04-08
Adjusted C13.2.8 Level, fixed a typo in C14.1.2

rakeshcorp:
Changes:

1. Level Adjustment: Move C13.2.8 (Session-level conversation trajectory analysis) from Level 3 to Level 2. Multi-turn jailbreaks are no longer "advanced" attacks; they are the baseline for modern jailbreak frameworks.
2. Fixed a typo C14.1.2
3. Added my name in the contributor section


RicoKomenda:
Hi @rakeshcorp,

thanks for the PR.

You only did the changes in the wiki, not the control list.

Also keep in mind, that there are multiple jailbreak controls in the AISVS. This one is about logging and monitoring. And this one especially, that it get's detection abilities with a multi-turn jailbreak pattern attack. In my opinion, most logging and monitoring tools for input to AI models are still pretty use-case-based hard to implement.

Also, please don't add yourself as contributor yet. We’re still trying to figure out exactly when someone is considered a contributor - we’ll also be updating the contributor list before the 1.0 release; see #102. For me, simply making a change to the wiki wouldn’t be enough.

I don’t want to discourage you. We’re still looking for contributions and adjustments to the current controls throughout April (control freeze at the end of April). As things stand, I’d close your PR in the next few days, as I still consider it a Level 3 issue at this stage, unless you can provide me with some good counterarguments.

PR #663, opened 2026-04-09
Add C9.7.6: prevent agent output from authorizing external state writes

RicoKomenda:
## Summary

- Adds control **9.7.6** (Level 2) to C9.7 Intent Verification and Constraint Gates
- Closes the gap where no existing control explicitly prevented agent-generated outputs from serving as the sole authorization signal for writes to persistent external state (file systems, databases, code repositories, configuration stores, external APIs)

## Problem

Prompt injection can achieve persistence when an injected instruction in a tool output causes the agent to write to external state (a config file, a database record, a code repository) that is later re-read. The injection survives the session because the write committed it.

Existing controls do not cover this:

- **C7.4.2** requires user confirmation for high-impact actions broadly, but does not address the output-as-authorization pattern
- **C9.7.3** detects unintended side effects post-execution, detection only
- **C8.2.4** covers the same loop for vector/memory only
- **C9.6.1** enforces authorization policy but assumes the authorization check is independent of model output -- this control makes that assumption explicit for the write case

## New control

> **9.7.6** | Verify that agent-generated outputs cannot serve as the sole authorization signal or parameter source for write operations to persistent external state (e.g., file systems, databases, code repositories, configuration stores, or external APIs), and that such writes require either explicit human authorization or an independent policy-based authorization check that evaluates the write against the original user intent rather than the model's output. | 2

### DIFF
--- a/1.0/en/0x10-C09-Orchestration-and-Agentic-Action.md
+++ b/1.0/en/0x10-C09-Orchestration-and-Agentic-Action.md
@@ -110,7 +110,7 @@ Prevent "technically authorized but unintended" actions by binding execution to 
 | **9.7.3** | **Verify that** post-execution checks detect unintended side effects. | 2 |
 | **9.7.4** | **Verify that** any mismatch between intended outcome and actual results triggers containment and, where supported, compensating actions. | 2 |
 | **9.7.5** | **Verify that** prompt templates and agent policy configurations retrieved from a remote source are integrity-verified at load time against their approved versions (e.g., via hashes or signatures). | 3 |
-| **9.7.6** | **Verify that** agent-generated outputs cannot serve as the sole authorization signal or parameter source for write operations to persistent external state (e.g., file systems, databases, code repositories, configuration stores, or external APIs), and that such writes require either explicit human authorization or an independent policy-based authorization check that evaluates the write against the original user intent rather than the model's output. | 2 |
+| **9.7.6** | **Verify that** all write operations to persistent external state are authorized by either explicit human approval or an independent policy-based authorization mechanism that evaluates the operation against the original user intent, and not solely on agent-generated output. | 2 |
 
 ---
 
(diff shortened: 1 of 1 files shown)


emmanuelgjr:
Strong support for this proposal. This addresses a critical gap in the persistence attack surface for agentic systems.

**From a data security perspective**, the risk here is that prompt injection can escalate from a transient context manipulation to a durable data integrity compromise. Once an agent writes to persistent external state (databases, file systems, repos, APIs), the attacker achieves:

1. **Persistence beyond session scope** — the injected action survives conversation reset
2. **Lateral impact** — other users/agents reading from the compromised store inherit the poisoned state
3. **Audit trail pollution** — the write appears to originate from an authorized agent, not the attacker

A few suggestions to strengthen the control:

**Scope clarification:** The current wording "persistent external state" could benefit from examples in the description. I'd suggest explicitly listing: databases, file systems, version control repos, cloud storage, external APIs with side effects, and configuration stores. This helps auditors scope their testing.

**Relationship to C8.2.4:** This control and 8.2.4 ("agent's own outputs are not automatically written back into its trusted memory") form a complementary pair — 8.2.4 covers internal memory, 9.7.6 covers external state. A cross-reference between them would help implementers see the full picture.

**Testability:** An auditor could verify this by: (1) injecting a prompt that instructs the agent to write to an external store, (2) confirming that the write is either blocked or requires independent authorization (human approval, separate policy evaluation, or secondary verification). This is independently testable as written.

Level 1 assignment is appropriate — this is a baseline defense against a well-documented attack vector.

1.0/en/0x10-C09-Orchestration-and-Agentic-Action.md
```diff
@@ -110,6 +110,7 @@ Prevent "technically authorized but unintended" actions by binding execution to
 | **9.7.3** | **Verify that** post-execution checks detect unintended side effects. | 2 |
 | **9.7.4** | **Verify that** any mismatch between intended outcome and actual results triggers containment and, where supported, compensating actions. | 2 |
 | **9.7.5** | **Verify that** prompt templates and agent policy configurations retrieved from a remote source are integrity-verified at load time against their approved versions (e.g., via hashes or signatures). | 3 |
+| **9.7.6** | **Verify that** agent-generated outputs cannot serve as the sole authorization signal or parameter source for write operations to persistent external state (e.g., file systems, databases, code repositories, configuration stores, or external APIs), and that such writes require either explicit human authorization or an independent policy-based authorization check that evaluates the write against the original user intent rather than the model's output. | 2 |
```
jmanico: _Verify that all write operations to persistent external state are authorized by either explicit human approval or an independent policy-based authorization mechanism that evaluates the operation against the original user intent, and not solely on agent-generated output._

ottosulin (approved):

PR #683, opened 2026-04-13
Add C9.7.7: Fail-closed on governance PDP unavailability

### LINKED ISSUE #676: Update C09 - Add fail-closed requirement when governance evaluation is unavailable
C9.7.1 requires pre-execution gates that block execution on policy violations, and C5.3.2 requires that failed authorization evaluations abort queries. These handle the case where the policy decision point returns a denial.
Neither addresses the case where the policy decision point is unreachable — network partition, service timeout, component failure. In this scenario the agent receives no authorization decision at all. Without an explicit fail-closed requirement for governance unavailability, implementations may default to fail-open (permit execution when no decision is received), creating a structural bypass: an agent or attacker that can disrupt communication with the governance layer effectively disables governance.

This is distinct from C9.3.2 (tool quota breaches fail closed) which addresses resource limits within sandboxes, not governance system availability.
Proposed control (9.7.5):

| # | Description | Level | Role |
| --- | --- | --- | --- |
| **9.7.5** | **Verify that** if the policy decision point is unreachable or fails to return an authorization decision within a defined timeout, execution is blocked by default and the failure is logged with sufficient context for incident investigation. | 2 | D/V |

This ensures that governance unavailability produces the same operational result as an explicit denial rather than permitting ungoverned execution. It aligns with the NIST SP 800-207 zero trust principle that access is denied by default when policy cannot be evaluated.

emmanuelgjr:
## Summary
- Adds **C9.7.7** (Level 2): requires agent execution to fail closed when the policy decision point (PDP) is unavailable (timeout, network partition, service failure), blocking the proposed action and logging the event

## Rationale
C5.3.2 covers failed authorization evaluations (PDP returning denial) and C9.7.1 covers violations returned by the gate, but neither explicitly addresses the case where no response arrives from the PDP. This control closes that gap by requiring fail-closed behavior on PDP unavailability.

Testable: simulate PDP unavailability, confirm execution is blocked, confirm the event is logged.

See discussion in #676 for full analysis.

## Test plan
- [ ] Verify C9.7.7 numbering does not conflict (9.7.5 and 9.7.6 are taken)
- [ ] Confirm Level 2 assignment is appropriate
- [ ] Verify control is testable as described

Closes #676

### DIFF
--- a/1.0/en/0x10-C09-Orchestration-and-Agentic-Action.md
+++ b/1.0/en/0x10-C09-Orchestration-and-Agentic-Action.md
@@ -111,6 +111,7 @@ Prevent "technically authorized but unintended" actions by binding execution to 
 | **9.7.4** | **Verify that** any mismatch between intended outcome and actual results triggers containment and, where supported, compensating actions. | 2 |
 | **9.7.5** | **Verify that** prompt templates and agent policy configurations retrieved from a remote source are integrity-verified at load time against their approved versions (e.g., via hashes or signatures). | 3 |
 | **9.7.6** | **Verify that** all write operations to persistent external state are authorized by either explicit human approval or an independent policy-based authorization mechanism that evaluates the operation against the original user intent, and not solely on agent-generated output. | 2 |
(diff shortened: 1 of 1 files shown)


RicoKomenda (commented):
Numbering is correct (9.7.5 and 9.7.6 are both taken), the gap analysis from #676 holds, and the wording is clean. "Fails closed by blocking the proposed action" is unambiguous and testable by simulating PDP unavailability. Level 2 is right.

Approve.

PR #685, opened 2026-04-13
Add scope note to C8.1: RAG document ACL inheritance

emmanuelgjr:
## Summary
- Adds a **scope note** to C8.1 section header clarifying that the RAG chunking/embedding pipeline is an ETL process for the purposes of C5.3.6
- Establishes that source document access policies (permitted roles, tenants) must propagate to derived chunks during ingestion
- Specifies that chunks without determinable source ACL must be treated as maximally restricted by default

## Rationale
C5.3.6 already requires row-level security and field-level masking policies to be inherited by derived data stores and not bypassed by ETL processes. The RAG ingestion pipeline maps directly onto this, but no auditor would test a RAG pipeline against C5.3.6 without an explicit cross-reference. A scope note makes the existing coverage discoverable without creating a duplicate control.

Per reviewer consensus in #667 (Rico + Otto), a scope note is preferred over a new standalone control.

## Test plan
- [ ] Verify the scope note accurately describes C5.3.6's applicability
- [ ] Confirm no existing scope note conflicts

Closes #667

PR #686, opened 2026-04-13
Strengthen C5.6.4: Runtime isolation for agent PDP

### LINKED ISSUE #662: C09 - Add independent governance authority requirement to C9.7
C9.7.1 requires pre-execution gates that evaluate proposed actions against hard policy constraints. This is a strong baseline, but it does not require that the evaluating component is architecturally independent from the orchestrator and executor.
The Anthropic Claude Mythos Preview system card (April 7, 2026) documented a frontier model that escaped its sandbox, sent unauthorized communications, manipulated its own git history, and engaged in covert reasoning to evade monitoring — all because policy evaluation operated within the same execution environment the model controlled. Framework-native guardrails that share the agent’s execution surface are structurally bypassable.
NIST SP 800-207 mandates logical separation of control plane and data plane. The OWASP State of Agentic AI Security and Governance v2 (April 2026) documents that deterministic hook points within orchestration frameworks “function more reliably as an early warning layer than as a hard security boundary.

Proposed control (C9.7.5):
        Description:
9.7.5.  Verify that autonomous agent actions are evaluated by a governance consultant component that is architecturally independent from the executing components, operates against a versioned policy set, and produces a verifiable authentication determination before any execution is permitted.  

This extends 9.7.1 by requiring that the evaluation function is not embedded in the orchestration runtime where the agent can influence, simulate, or bypass it. It complements C5.2.6 (externalized PDP) by making architectural independence a requirement specifically for autonomous agent action governance, not just traditional access control.
Level 2 is appropriate because production agentic systems handling sensitive data require this separation. Level 3 environments (safety-critical, regulated) could mandate additional properties on top of this baseline.

emmanuelgjr:
## Summary
- **Updates C5.6.4** to explicitly require that the policy decision point is isolated from the agent's execution environment, so a compromised agent cannot influence or bypass evaluation
- Adds requirement that the PDP receives structured action descriptions rather than the agent's raw reasoning context
- Adds **scope note** cross-referencing C14.1, C14.2, and C9.6.4 for the escalation path

## Rationale
Existing controls (C5.2.6, C5.6.4, C9.6.4) already require an external PDP for agent authorization, but none explicitly state that the PDP must be isolated at runtime such that a compromised agent cannot reach or influence it. The data-plane isolation concern (PDP should not ingest the raw agent context that may contain injected content) was raised in the discussion and is addressed in the scope note.

Per reviewer consensus in #662 (Rico + Otto), strengthening C5.6.4 with a scope note is preferred over a new standalone control.

## Test plan
- [ ] Verify updated C5.6.4 wording is clear and auditable
- [ ] Confirm cross-references to C14.1, C14.2, C9.6.4 are accurate
- [ ] Verify Level 3 assignment remains appropriate

Closes #662

### DIFF
--- a/1.0/en/0x10-C05-Access-Control-and-Identity.md
+++ b/1.0/en/0x10-C05-Access-Control-and-Identity.md
@@ -87,7 +87,9 @@ Control permissions for AI agents and autonomous systems through scoped capabili
 | **5.6.1** | **Verify that** autonomous agents receive scoped capability tokens that explicitly enumerate permitted actions, accessible resources, time boundaries, and operational constraints. | 1 |
 | **5.6.2** | **Verify that** high-risk capabilities (file system access, code execution, external API calls, financial transactions) are disabled by default and require explicit authorization. | 1 |
 | **5.6.3** | **Verify that** capability tokens are bound to user sessions, include cryptographic integrity protection, and cannot be persisted or reused across sessions. | 2 |
-| **5.6.4** | **Verify that** agent-initiated actions undergo authorization through a policy decision point that evaluates contextual attributes (e.g., user identity, resource sensitivity, action type, environmental context). | 3 |
+| **5.6.4** | **Verify that** agent-initiated actions undergo authorization through a policy decision point that is isolated from the agent's execution environment such that a compromised or manipulated agent runtime cannot influence or bypass the evaluation, and that evaluates contextual attributes (e.g., user identity, resource sensitivity, action type, environmental context) based on a structured action description rather than the agent's raw reasoning context. | 3 |
+
+> **Scope note -- C5.6.4 runtime isolation.** The policy decision point must receive a sanitized, structured representation of the proposed action (action type, target resource, parameters) rather than the agent's full context window, which may contain injected content that could influence the evaluation. This isolation requirement complements C14.1 (human oversight gates) and C9.6.4 (model cannot make access control decisions). For high-risk actions, C14.2 defines the escalation path from automated governance evaluation to human approval.
 
 ---
 
--- a/1.0/en/0x10-C05-Access-Control-and-Identity.md
+++ b/1.0/en/0x10-C05-Access-Control-and-Identity.md
@@ -82,6 +82,8 @@ Ensure logical and cryptographic isolation between tenants in shared AI infrastr
 
 Control permissions for AI agents and autonomous systems through scoped capability tokens and continuous authorization.
 
+> **Scope note:** Runtime isolation of the policy decision point from the agent execution environment complements C9.6.4 (model cannot make access control decisions) and C14.2 (human oversight escalation for high-risk actions).
+
 | # | Description | Level |
 | :--------: | --------------------------------------------------------------------------------------------- | :---: |
 | **5.6.1** | **Verify that** autonomous agents receive scoped capability tokens that explicitly enumerate permitted actions, accessible resources, time boundaries, and operational constraints. | 1 |
--- a/1.0/en/0x10-C05-Access-Control-and-Identity.md
+++ b/1.0/en/0x10-C05-Access-Control-and-Identity.md
@@ -82,14 +82,14 @@ Ensure logical and cryptographic isolation between tenants in shared AI infrastr
 
 Control permissions for AI agents and autonomous systems through scoped capability tokens and continuous authorization.
 
+> **Scope note:** The policy decision point must receive a sanitized, structured representation of the proposed action (action type, target resource, parameters) rather than the agent's full context window, which may contain injected content that could influence the evaluation. This isolation requirement complements C14.1 (human oversight gates) and C9.6.4 (model cannot make access control decisions). For high-risk actions, C14.2 defines the escalation path from automated governance evaluation to human approval.
+
 | # | Description | Level |
 | :--------: | --------------------------------------------------------------------------------------------- | :---: |
 | **5.6.1** | **Verify that** autonomous agents receive scoped capability tokens that explicitly enumerate permitted actions, accessible resources, time boundaries, and operational constraints. | 1 |
 | **5.6.2** | **Verify that** high-risk capabilities (file system access, code execution, external API calls, financial transactions) are disabled by default and require explicit authorization. | 1 |
 | **5.6.3** | **Verify that** capability tokens are bound to user sessions, include cryptographic integrity protection, and cannot be persisted or reused across sessions. | 2 |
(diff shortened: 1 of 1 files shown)


RicoKomenda:
Scope notes are always on top of the table, starting right after section title

jmanico:
1. This is a compound requirement that we wish to avoid. The rewritten 5.6.4 now bundles three distinct verification items:
  - (a) PDP exists and evaluates contextual attributes (original)
  - (b) PDP is runtime-isolated from agent execution environment (new)
  - (c) PDP receives structured action descriptions, not raw agent context (new)

  An auditor could pass (a) and (c) but fail (b). Per project guidelines, these should be separate requirements.

  2. The requirement is now a wall of text. It's 52 words in a single sentence. That's hard to audit against and hard to read in the table. Compare with the clean, scannable requirements elsewhere in C5.6.

  3. Scope note duplicates the requirement. The scope note restates the structured-input requirement verbatim ("must receive a sanitized, structured representation of the proposed action (action type, target resource, parameters) rather than the agent's full context window"). The scope note should cross-reference and clarify boundaries, not repeat what the requirement row already says.

  4. Scope note reads as prescriptive (same pattern as PR #685). "The policy decision point must receive..." is requirement language, not scope guidance.

jmanico:
It we really want to keep these, then I would add these before Otto's reduction pass of C5.

* 5.6.4 (L3): Verify that agent-initiated actions undergo authorization through a policy decision point that evaluates contextual attributes (e.g., user identity, resource sensitivity, action type, environmental context).
* 5.6.5 (L3): Verify that the policy decision point for agent authorization is isolated from the agent's execution environment such that a compromised agent runtime cannot influence or bypass evaluation.
* 5.6.6 (L3): Verify that the policy decision point receives structured action descriptions (action type, target resource, parameters) rather than the agent's raw reasoning context.

1.0/en/0x10-C05-Access-Control-and-Identity.md
```diff
@@ -82,12 +82,14 @@ Ensure logical and cryptographic isolation between tenants in shared AI infrastr
 
 Control permissions for AI agents and autonomous systems through scoped capability tokens and continuous authorization.
 
+> **Scope note:** The policy decision point must receive a sanitized, structured representation of the proposed action (action type, target resource, parameters) rather than the agent's full context window, which may contain injected content that could influence the evaluation. This isolation requirement complements C14.1 (human oversight gates) and C9.6.4 (model cannot make access control decisions). For high-risk actions, C14.2 defines the escalation path from automated governance evaluation to human approval.
```
ottosulin: Two issues with the scope note:

1. Prescriptive language. The scope note says "The policy decision point must receive a sanitized, structured representation..." - that is requirement language, not scope guidance. Scope notes in this project clarify boundaries, cross-reference related controls, and explain what is/isn't in scope. They do not introduce new "must" obligations. Compare with the C14.2 scope note, which says "Compliance with C14.2 requires demonstrating that a documented policy exists, not merely that approvals occur" - that is boundary clarification, not a new requirement.
2. Duplicates the requirement. The structured-input requirement appears both in the scope note and in the rewritten 5.6.4 row. If we keep both, an auditor reads the same thing twice.

If we split into separate requirements (see below), the scope note should be reduced to just the cross-references: something like "Runtime isolation of the policy decision point from the agent execution environment complements C9.6.4 (model cannot make access control decisions) and C14.2 (human oversight escalation for high-risk actions)."

1.0/en/0x10-C05-Access-Control-and-Identity.md
```diff
@@ -82,12 +82,14 @@ Ensure logical and cryptographic isolation between tenants in shared AI infrastr
...
 | **5.6.2** | **Verify that** high-risk capabilities (file system access, code execution, external API calls, financial transactions) are disabled by default and require explicit authorization. | 1 |
 | **5.6.3** | **Verify that** capability tokens are bound to user sessions, include cryptographic integrity protection, and cannot be persisted or reused across sessions. | 2 |
-| **5.6.4** | **Verify that** agent-initiated actions undergo authorization through a policy decision point that evaluates contextual attributes (e.g., user identity, resource sensitivity, action type, environmental context). | 3 |
+| **5.6.4** | **Verify that** agent-initiated actions undergo authorization through a policy decision point that is isolated from the agent's execution environment such that a compromised or manipulated agent runtime cannot influence or bypass the evaluation, and that evaluates contextual attributes (e.g., user identity, resource sensitivity, action type, environmental context) based on a structured action description rather than the agent's raw reasoning context. | 3 |
```
ottosulin: Compound requirement. The rewritten 5.6.4 now bundles three independently testable verification items:

- (a) PDP exists and evaluates contextual attributes (the original requirement)
- (b) PDP is runtime-isolated from the agent execution environment
- (c) PDP receives structured action descriptions, not raw agent context

An auditor could pass (a) and (c) but fail (b). Per project guidelines, each requirement should be independently assessable.

Also, at 52 words in a single sentence, this is hard to read and audit against. Compare with the clean, scannable requirements in the rest of C5.6.

Suggested approach (aligning with @jmanico's proposal): split into three separate requirements:

- 5.6.4 (L3): Keep the original wording - "Verify that agent-initiated actions undergo authorization through a policy decision point that evaluates contextual attributes (e.g., user identity, resource sensitivity, action type, environmental context)."
- 5.6.5 (L3): "Verify that the policy decision point for agent authorization is isolated from the agent's execution environment such that a compromised agent runtime cannot influence or bypass evaluation."
- 5.6.6 (L3): "Verify that the policy decision point receives structured action descriptions (e.g., action type, target resource, parameters) rather than the agent's raw reasoning context."

All three at L3 seems right: 5.6.4 was already L3, and both new controls require architectural decisions that go beyond what most systems do today. They are important for high-assurance environments but not baseline.

ottosulin (changes requested):
Thanks for following through on the #662 discussion, @emmanuelgjr. The underlying concern (PDP runtime isolation from agent execution environment) is real and worth including. However, the current approach has a few issues that I think need to be addressed before merging.

I agree with @jmanico's feedback: the rewritten 5.6.4 is now a compound requirement, and the scope note contains prescriptive language that duplicates the requirement text. I would recommend the split approach Jim suggested. Added more detailed comments to help with this.

And yes ideally let's merge this if possible before C05 reduction pass. I may do C06 while waiting for this to get sorted out.

ottosulin (approved):
Thanks @RicoKomenda looks good now!

PR #689, opened 2026-04-13
Add architectural note to C9.6: evaluation-only policy components

### LINKED ISSUE #675: Update C09 - Add non-execution constraint for governance evaluation components
C5.6.4 (with the proposed runtime isolation strengthening in PR from #662) requires that the policy decision point is isolated from the agent's execution environment. C9.6.4 requires that access control decisions are never made by the AI model itself.
These controls establish that the evaluator is separate from the agent and that the model cannot make access control decisions. However, neither addresses the execution capabilities of the governance component itself. A policy decision point that is isolated from the agent runtime but retains the ability to execute actions in the operational environment (e.g., terminate processes, modify configurations, write to production datastores) introduces a different risk: the governance component becomes an execution surface that can be targeted, co-opted, or exploited as an alternative pathway for unauthorized actions.

NIST SP 800-207 separates the control plane from the data plane and constrains each to its designated function. The same principle applied to agent governance means the evaluation component should be constrained to evaluation — it should not possess operational execution capabilities that could be exploited if the component is compromised.
Proposed control (9.6.5):

| # | Description | Level | Role |
| --- | --- | --- | --- |
| **9.6.5** | **Verify that** components serving as the policy decision point for autonomous agent actions are constrained to evaluation and decision functions and do not possess the ability to execute actions in the operational environment (e.g., invoke tools, modify resources, initiate transactions, or issue commands to downstream systems). | 3 | D/V |

This complements C9.6.4 (model cannot make access control decisions) by constraining the governance component from the other direction: just as the model must not evaluate policy, the policy evaluator must not execute actions. This bidirectional separation reduces the attack surface of the governance layer and prevents it from being weaponized as an execution pathway.

emmanuelgjr:
## Summary
- Adds **architectural note** in C9.6 clarifying that policy decision components for agent authorization should be evaluation-only: they evaluate proposed actions and return permit/deny decisions, but do not execute actions, invoke tools, or modify resources
- References NIST SP 800-207 control-plane/data-plane separation principle

## Rationale
The original proposal (#675) suggested a standalone verifiable control requiring governance components to not possess execution ability. Reviewer feedback identified two issues: (1) the constraint is not clearly verifiable as a standalone control (an auditor cannot reliably enumerate all execution pathways), and (2) it is not AI-specific — control-plane/data-plane separation is a general NIST SP 800-207 principle.

The existing C9.6.4, C5.6.4, and C5.2.6 already push implementers toward purpose-built evaluation components. An architectural note makes this design expectation explicit without creating a control that is hard to audit independently.

## Test plan
- [ ] Verify the architectural note is consistent with C9.6.4, C5.6.4, C5.2.6
- [ ] Confirm NIST SP 800-207 reference is appropriate

Closes #675

### DIFF
--- a/1.0/en/0x10-C09-Orchestration-and-Agentic-Action.md
+++ b/1.0/en/0x10-C09-Orchestration-and-Agentic-Action.md
@@ -87,6 +87,8 @@ Protect agent-to-agent and agent-to-tool communications from hijacking, injectio
 
 Ensure every action is authorized at execution time and constrained by scope.
 
+> **Architectural note -- evaluation-only policy components.** Policy decision components used for agent authorization (C9.6.4, C5.6.4, C5.2.6) should be evaluation-only: they evaluate proposed actions against policy rules and return permit/deny decisions, but do not themselves execute actions, invoke tools, or modify external resources. This separation follows the NIST SP 800-207 control-plane/data-plane principle. Standard purpose-built policy engines (e.g., OPA, Cedar) are evaluation-only by design.
+
 | # | Description | Level |
 | :--: | --- | :---: |
 | **9.6.1** | **Verify that** agent actions are authorized against fine-grained policies enforced by the runtime that restrict which tools an agent may invoke, which parameter values it may supply (e.g., allowed resources, data scopes, action types), and that policy violations are blocked. | 1 |


RicoKomenda:
After thinking about it... @ottosulin / @jmanico do we need an architectural note for a section? Or would that be too much for the people using the AISVS?

RicoKomenda:
I think we should keep an eye out for a scope note rather than an architectural note, even if it was my first intent to do that. For now, I would close this PR and this issue, and we'll see if this reqs are still there after the reduction pass of C9.

PR #704, opened 2026-04-17
C02 - Reduction

DotDotSlashRepo:
Removed 1 control from "C2.2 Pre-Tokenization Input Normalization"

* Control 2.2.2 ("Verify that suspected adversarial inputs are quarantined and logged") duplicates monitoring/logging concerns that are already comprehensively covered in C13 (Monitoring, Logging & Anomaly Detection)
* The logging and quarantine function is a cross-cutting control that applies across inputs in all stages
* Keeping it in C2.2 creates unnecessary redundancy when C13 already specifies detailed logging requirements (13.1.1-13.1.8) and abuse detection/alerting (13.2.1-13.2.10)

### DIFF
--- a/1.0/en/0x10-C02-User-Input-Validation.md
+++ b/1.0/en/0x10-C02-User-Input-Validation.md
@@ -33,8 +33,7 @@ AI models process text through tokenizers and embeddings that can be exploited v
 | # | Description | Level |
 | :--------: | ------------------------------------------------------------------------------------------------------------------- | :---: |
 | **2.2.1** | **Verify that** input normalization (Unicode NFC canonicalization, homoglyph mapping, removal of control and invisible Unicode characters, and bidirectional text neutralization) is applied before tokenization or embedding, and that inputs which still contain suspicious encoding artifacts after normalization are rejected or flagged for review. | 1 |
-| **2.2.2** | **Verify that** suspected adversarial inputs are quarantined and logged. | 1 |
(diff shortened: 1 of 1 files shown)


RicoKomenda (commented):
The logging argument holds - C13.1 and C13.2 cover detection, alerting, and audit logging comprehensively. Removing the "logged" obligation from C2 is justified.

But "quarantined" is not the same as "alerted." C13.2.1 requires detection and alerting. It does not require blocking the input. C2.1.1 gates inputs that fail prompt injection screening, and C2.2.1 rejects encoding artifacts, but neither is a general-purpose gate for any adversarial detection result. Removing 2.2.2 creates a real gap: there is no longer an explicit requirement to stop a detected adversarial input from proceeding into the prompt or triggering an action.

One option to close that gap is to keep the control but tighten it to just the gate action:

> Verify that inputs identified as adversarial by any detection mechanism are blocked from inclusion in prompts or execution of actions.

That drops the "logged" clause (already in C13) and makes the requirement more concrete. Alternatively, if the view is that C2.1.1 is sufficient as the primary gate, a scope note there explaining the relationship would help.

DotDotSlashRepo:
I think "Verify that inputs identified as adversarial by any detection mechanism are blocked from inclusion in prompts or execution of actions." is more specific and guides action. Shall I update this ?

RicoKomenda:
Yes, please update with that wording. Rather than adding a new control, revert the removal of 2.2.2 and replace the existing text with the tighter wording:

> Verify that inputs identified as adversarial by any detection mechanism are blocked from inclusion in prompts or execution of actions.

That keeps the numbering clean (2.2.1 normalization, 2.2.2 gate) and drops the 'logged' clause without leaving a gap on the blocking side.

jmanico (approved):

PR #680, opened 2026-04-12
Reduction pass C03

ottosulin:
Polishes C03 (Model Lifecycle Management) by removing or reworking requirements that are not AI-specific or duplicate controls already present in ASVS v5 or other AISVS chapters (C1, C4).

### Removed: 3.4.1 (environment separation)

Generic DevOps/security practice ("dev, testing, and production environments are logically or physically separated with distinct access controls"). Not AI-specific - applies equally to any software system. Already covered by:
- AISVS C4.3.2: "AI workloads across environments run in isolated network segments with no direct internet access and no cross-environment network connectivity."
- AISVS C4.3.6: "environments use separate identity roles and security groups with no shared principals or credentials across environment boundaries."

### Removed: 3.4.5 (training data source validation)

Heavily overlaps with C1 (Training Data Integrity & Traceability):
- C1.1.1 covers inventory of training data sources with origin and processing history.
- C1.2.4 covers cryptographic hashes/signatures for data integrity.
- C1.2.5 covers automated integrity monitoring against unauthorized modifications.
- C1.5.1 covers lineage of each dataset including transformations.

The tail clause about "RAG indexes, tool logs, and agent-generated data used for fine-tuning" is already scoped into C1 by its general applicability to all training data.

### Reworked: 3.4.2 (environment isolation)

The original mixed generic infrastructure isolation (already in C4.3.2) with AI-specific concerns. Reworked to focus only on the AI-specific layer: agent orchestration runtimes, tool/MCP servers, and model registries must not be shared across environment boundaries. Removed the generic "infrastructure, data stores" part that C4 already covers.

### Reworked: 3.4.3 (version control and peer review)

The original was a compound requirement mixing a generic software practice (use version control + peer review) with AI-specific artifact types. "Stored in version control and require peer review" is standard SDLC, not AI-specific. Reworked to focus on the AI-specific value: ensuring that AI-specific configuration artifacts (prompt templates, agent policies, tool/MCP schemas, capability allow-lists) are subject to the same version control and review rigor as application code. This framing acknowledges these artifact types are often treated informally and highlights the AI-specific gap.

### Reworked: 3.4.4 (training environment isolation)

The original overlapped with C4.3.1 (default-deny network policies) and C4.3.2 (isolated network segments). Reworked to focus on the AI-specific isolation concern: training and fine-tuning environments must not have access to production model endpoints, agent orchestration services, tool/MCP servers, or live RAG data sources. The generic network isolation is left to C4.

## Requirements retained without changes

[...]

PR #707, opened 2026-04-18
C03 Fixes and Reduction

DotDotSlashRepo:
* Formatting edits on introduction paragraph
* Merged 3.2.2 to 3.2.1 because of similarity
* Made edits to improve grammar of 3.2.6
* Added reference link to OWASP AITG

### DIFF
--- a/1.0/en/0x10-C03-Model-Lifecycle-Management.md
+++ b/1.0/en/0x10-C03-Model-Lifecycle-Management.md
@@ -28,12 +28,14 @@ Models must pass defined security and safety validations before deployment.
 
 | # | Description | Level |
 | :--------: | ------------------------------------------------------------------------------------------------------------------- | :---: |
-| **3.2.1** | **Verify that** models undergo automated input validation testing and output sanitization testing before deployment. | 1 |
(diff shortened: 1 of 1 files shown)


RicoKomenda (commented):
Thanks for the fixes. A few things to note before this can merge.

**The merge of 3.2.1 and 3.2.2 should be reverted.** PR #680 (already merged) deliberately split the old compound control into two separate L1 requirements because input validation testing and output sanitization testing are independently verifiable. A system could pass one and fail the other. Re-merging them undoes that intentional split and reintroduces a compound control. Please keep 3.2.1 and 3.2.2 as separate controls and restore the original numbering through 3.2.7.

**Appendix D** would also need to be updated if the renumbering stands, which is another reason to avoid it.

**The other changes are all good** and should be kept: the em-dash removal in the control objective, the grammar fix on 3.2.6 ("is provided" added), and the OWASP AI Testing Guide reference.

RicoKomenda:
Hi @DotDotSlashRepo: Can you change the proposed fixes from earlier?

DotDotSlashRepo:
Done changes as per comment

jmanico (approved):

PR #711, opened 2026-04-24
Update 0x10-C11-Adversarial-Robustness.md

sydseter:
Renumbering 11.5


sydseter:
As I see, some work is still ongoing on 11.5; perhaps it's too soon.

Issue #693, opened 2026-04-14
Audit: split compound requirements and reframe negative requirements


jmanico:
## Overview

A full scan of all 513 AISVS requirements identified **22 compound requirements** that bundle multiple independently-assessable verification items, and **8 negative requirements** where reframing positively would improve auditability.

Per project guidelines: "Each requirement must verify exactly one thing. Watch for 'and' or comma-separated clauses that smuggle in multiple verification items. If a requirement asks an auditor to check two or more distinct conditions, propose splitting it into separate requirements -- even if the conditions are related."

These can be addressed incrementally in per-chapter PRs. This issue serves as a tracker.

### Filtering criteria applied

- **Not compound:** A single action applied to a list of targets or fields (e.g., "log timestamp, user ID, session ID" is one logging action). Lists behind "e.g." are examples, not separate verification items.
- **Not negative:** Standard security verbs like "prevent unauthorized access" or "fail closed" are acceptable when the primary framing is positive. Only flagged when the entire core framing is negative.
- **Included:** Only requirements where an auditor could pass one part and fail another (compound), or where positive reframing would make the requirement clearer and more testable (negative).

---

## Part 1: Compound Requirements (split candidates)

Each of these bundles two or more verification items that an auditor could independently pass or fail.

### C02 -- User Input Validation

**2.1.2** -- Instruction hierarchy enforcement AND preservation across multi-step/tool-augmented workflows AND prompt composition constraint.
- Current text includes three distinct verification concerns joined across two sentences.
- Suggested split: (a) enforce instruction hierarchy where system/developer messages override user inputs, (b) hierarchy preserved across multi-step and tool-augmented workflows, (c) user-controlled content in prompt composition cannot override system instructions.

**2.2.1** -- Input normalization applied AND suspicious artifacts post-normalization rejected/flagged.
- Suggested split: (a) normalization (NFC, homoglyph, control chars, bidi) applied before tokenization, (b) post-normalization suspicious artifact rejection or flagging.

### C03 -- Model Lifecycle Management

**3.6.4** -- Stage output integrity-verified AND intermediate checkpoints registered as rollback artifacts.
- Suggested split: (a) inter-stage integrity verification, (b) checkpoint registration with per-stage rollback.

### C04 -- Infrastructure

**4.1.3** -- Read-only root filesystem AND writable mounts hardened with noexec/nosuid/nodev.
[...]

PR #773, opened 2026-05-09
fix(#693): split compound requirements in C02

RicoKomenda:
Splits two compound requirements in C2 Input Validation per #693.

**2.1.2** (hierarchy + multi-step + composition) → **2.1.2**, **2.1.5**
- 2.1.5 and 2.1.6 were initially split into two separate items; merged back into one per reviewer feedback (they verify the same underlying property end-to-end)
- "influence or override" narrowed to "cannot override" for auditability

**2.2.1** (normalization + artifact rejection) → **2.2.1**, **2.2.3**
- 2.2.3 placed at L2 per reviewer feedback (requires second-pass heuristic above L1 normalization)

Requirements sorted by level (L1 first, then L2, then L3) and renumbered sequentially.
Appendix D cross-references updated.

### DIFF
--- a/1.0/en/0x10-C02-Input-Validation.md
+++ b/1.0/en/0x10-C02-Input-Validation.md
@@ -15,14 +15,15 @@ Prompt injection is one of the top risks for AI systems. Defenses against this t
 | # | Description | Level |
 | :--------: | ------------------------------------------------------------------------------------------------------------------- | :---: |
 | **2.1.1** | **Verify that** all external or derived inputs that may steer model behavior are treated as untrusted and screened by a prompt injection detection ruleset or classifier before being included in prompts or used to trigger actions. | 1 |
-| **2.1.2** | **Verify that** the system enforces an instruction hierarchy in which system and developer messages override user instructions and other untrusted inputs, even after processing user instructions. This enforcement must be preserved across multi-step interactions and tool-augmented workflows. In such cases, prompt composition or intermediate outputs must not allow user-controlled content to influence or override system or developer instructions. | 1 |
+| **2.1.2** | **Verify that** the system enforces an instruction hierarchy in which system and developer messages override user instructions and other untrusted inputs, even after processing user instructions. | 1 |
 | **2.1.3** | **Verify that** input length controls prevent user-supplied content from exceeding a defined proportion of the context window, and that inputs exceeding token limits are rejected rather than silently truncated, ensuring system instructions and safety directives are not displaced from the model's effective attention. | 1 |
 | **2.1.4** | **Verify that** the system implements a character set limitation for user inputs to model prompts, allowing only characters that are explicitly required for business purposes using an allow-list approach. | 1 |
-| **2.1.5** | **Verify that** prompts originating from third-party content (web pages, PDFs, emails) are sanitized in isolation (for example, stripping instruction-like directives and neutralizing HTML, Markdown, and script content) before being concatenated into the main prompt. | 2 |
-| **2.1.6** | **Verify that** the system enforces per-request limits on the number of user-supplied demonstrations included in a single context window. | 2 |
-| **2.1.7** | **Verify that** the system detects patterns indicative of systematic in-context behavioral override attempts consistent with many-shot jailbreaking. | 2 |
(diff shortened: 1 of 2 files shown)


1.0/en/0x10-C02-Input-Validation.md
```diff
@@ -15,14 +15,16 @@ Prompt injection is one of the top risks for AI systems. Defenses against this t
...
-| **2.1.5** | **Verify that** prompts originating from third-party content (web pages, PDFs, emails) are sanitized in isolation (for example, stripping instruction-like directives and neutralizing HTML, Markdown, and script content) before being concatenated into the main prompt. | 2 |
-| **2.1.6** | **Verify that** the system enforces per-request limits on the number of user-supplied demonstrations included in a single context window. | 2 |
-| **2.1.7** | **Verify that** the system detects patterns indicative of systematic in-context behavioral override attempts consistent with many-shot jailbreaking. | 2 |
-| **2.1.8** | **Verify that** detected in-context behavioral override attempts are classified and handled as prompt injection events. | 2 |
-| **2.1.9** | **Verify that** prompt injection screening respects user-specific policies (age and regional legal constraints) via attribute-based rules resolved at request time, including the role or permission level of the calling agent. | 2 |
+| **2.1.5** | **Verify that** instruction hierarchy enforcement is preserved across multi-step interactions and tool-augmented workflows. | 1 |
```
ottosulin: Optional - just thinking which is the right way to go and laying my thoughts here, you make the decision: 2.1.5 and 2.1.6 verify the same underlying property (instruction hierarchy holding under multi-step / tool-augmented composition) from two phrasings. An auditor who confirms 2.1.5 has effectively confirmed 2.1.6, and vice versa. Consider merging into one item:

> "Verify that the instruction hierarchy is preserved across multi-step interactions and tool-augmented workflows, including prompt composition and intermediate outputs, such that user-controlled content cannot override system or developer instructions."

This stays atomic (one verification: hierarchy holds end-to-end) while capturing both concerns from the original 2.1.2.

1.0/en/0x10-C02-Input-Validation.md
```diff
@@ -15,14 +15,16 @@ Prompt injection is one of the top risks for AI systems. Defenses against this t
...
+| **2.1.6** | **Verify that** prompt composition and intermediate outputs in multi-step or tool-augmented workflows do not allow user-controlled content to influence or override system or developer instructions. | 1 |
```
ottosulin: If kept separate (see comment above):

I suggest replacing "influence or override" with "override". "Influence" is not auditable in any practical sense, while "override" is verifiable through targeted red-team prompts and tool-output injection tests.

1.0/en/0x10-C02-Input-Validation.md
```diff
@@ -32,9 +34,10 @@ AI models process text through tokenizers and embeddings that can be exploited v
...
+| **2.2.1** | **Verify that** input normalization (Unicode NFC canonicalization, homoglyph mapping, removal of control and invisible Unicode characters, and bidirectional text neutralization) is applied before tokenization or embedding. | 1 |
 | **2.2.2** | **Verify that** inputs identified as adversarial by any detection mechanism are blocked from inclusion in prompts or execution of actions. | 1 |
-| **2.2.3** | **Verify that** encoding and representation smuggling in both inputs and outputs (e.g., invisible Unicode/control characters, homoglyph swaps, or mixed-direction text) are detected and mitigated. Approved mitigations include canonicalization, strict schema validation, policy-based rejection, or explicit marking. | 3 |
+| **2.2.3** | **Verify that** inputs which still contain suspicious encoding artifacts after normalization are rejected or flagged for review. | 1 |
```
ottosulin: Suggest L2 rather than L1 for this requirement.

Detecting "suspicious encoding artifacts after normalization" requires a second-pass heuristic or classifier on top of the L1 normalization in 2.2.1. 

There is also bit of overlap with 2.2.4 (L3), which covers encoding/representation smuggling detection and mitigation more broadly. Placing this requirement at L1 between an L1 normalization control and an L3 smuggling-mitigation control creates an unclear gradient. L2 would better reflect both the implementation effort and the level gradient.

ottosulin (changes requested):

jmanico (approved):

Issue #814, opened 2026-05-13
AISVS release practices


ottosulin:
Opening this up for discussion, was briefly discussed on project leads call earlier but left for May to sort out.

Why I think we need to discuss this topic is that in general similar standards are updated with relatively slow cadence - which is very good, because it gives stability.

AISVS is addressing technology and risks that are constantly changing. With AI, a major release every few years will not be often enough. We need to able to update AISVS more often than ASVS. To help users keep up with content possibly changing more often, we should define the release practices clearly. Below is my proposal what it could look like, easier to show a proposal.

Questions about the proposal:

- Is something not clear there? Is something important information missing?
- Should minor release allow messing around more freely in a single section (e.g. remove / add single section), while keeping the chapter & control objective untouched?

-----

## AISVS releases

AISVS follows semantic versioning in releases, with following conventions:
- A patch release (e.g. 1.0.1) will not contain material changes to requirement contents, but may update language or technical references in the requirements to match changes in technology or correct errors.
- A minor release (e.g. 1.1.0) will add or remove requirements, or include material changes (e.g. modifying content or level) to existing requirements to match changes in technology. A minor release will not modify control objectives or sections. A minor release will contain all patch release changes from previous minor versions.
- A major release (e.g. 2.0.0) will contain additions, removals or major changes to entire chapters or sections.

Patch and minor releases may happen often, unlike major releases. 

For referencing requirements across versions, refer to (How to Reference AISVS Requirements)[https://github.com/OWASP/AISVS#how-to-reference-aisvs-requirements].

ajayojha:
Here are my quick thoughts:

* I may be missing something. But I do not currently see a dedicated release strategy or versioning governance document for AISVS in the repository structure.
* AISVS should follow a more flexible and responsive release model. This is because the AI threat landscape changes much faster than traditional application security.
* I also think minor releases should have more flexibility than currently proposed. Minor releases should not be too restricted from updating section text or refining control objectives when needed.
* While drafting the release strategy, I would suggest keeping in mind that continuous and incremental releases can provide strong value for organizations adopting AISVS. This helps organizations keep up with evolving AI security risks without requiring them to independently track every emerging AI security issue themselves.
* For major releases, we should have a clearly defined plan and document it somewhere within the project repository or governance documentation.

jmanico:
Thanks Otto. This is exactly the kind of release policy we should lock down before 1.0.

[...]

PR #818, opened 2026-05-15
docs: add release policy (#814)

jmanico:
## Summary

Closes the thread on #814 by writing down the AISVS release policy so contributors, adopters, auditors, and tooling vendors have a single source of truth.

- Adds `RELEASE.md` covering the two-part versioning scheme (`v<MAJOR>.<MINOR>`), the scope rules for patch, minor, and major releases, parallel maintenance for the previous major line, and the referencing convention.
- Links the policy from `README.md` (Versioning section) and `CONTRIBUTING.md`.
- Captures the strict minor-release rule from the issue thread: no chapter, section, or control-objective changes in a minor release, since the `v<version>-Cx.y.z` referencing convention encodes chapter and section identity.

## Notes

- Two-part versioning is consistent with what is already in the repo (`1.0/`, `1.01-dev/`) and with the `v1.0-C9.4.3` citation format already documented in the README.
- Cadence is intentionally not pinned to a calendar. Releases ship when the content warrants it.
- Parallel maintenance is described as a commitment with the specific window announced at the time the next major opens, rather than fixing a number now.

## Test plan

- [ ] Review `RELEASE.md` against the discussion in #814 and confirm both Otto's original proposal and the strict-minor adjustment are reflected.
- [ ] Confirm README and CONTRIBUTING links resolve.
- [ ] Sanity-check the repository layout block in `RELEASE.md` against current folder names.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -56,4 +56,4 @@ We are looking for help with translations after v1.0 is released!
 
 ## Release Policy
 
-AISVS follows a published release policy that defines what can change in patch, minor, and major releases, and how versions are referenced across time. See [RELEASE.md](RELEASE.md) before proposing changes that add, remove, or restructure requirements.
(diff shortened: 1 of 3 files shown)


ottosulin (approved):
Exactly what we need and very clearly written. Nice work @jmanico!

Issue #793, opened 2026-05-10
Flag the compound requirements


vtknightmare:
These each verify more than one thing and an auditor could pass one part and fail another:

- C2.1.2 — instruction hierarchy + multi-step preservation + prompt composition rule
- C2.2.1 — NFC + homoglyph + control-char + bidi + reject/flag policy
- C2.2.3 — input AND output encoding smuggling (output side belongs in C7)
- C3.2.8 — also has a grammar issue: "unless an explicit override approval from pre-designated authorized personnel with documented business justifications" is missing a verb
- C7.6.2 — claim traceability AND unsupported-claim flagging
- C8.1.5 — three distinct anomaly indicators in one verify
- C12.5.1 — consent scope check + refuse-or-downgrade + scope-coverage / data-subjects
- C13.1.1 — long enumeration of logged fields
I'd flag these for splitting or rewording, not split them automatically — final 1.0 should preserve numbering as much as possible.

jmanico:
Thanks for the second pass on these. I went through each against current text and here's where I landed.

A couple have already been split since you scanned:
  - 2.1.2 now just covers the hierarchy itself, and the multi-step preservation piece is 2.1.5
  - 13.1.1 was split in #778 so the "exclude content from logs" half lives in 13.1.2

A couple I'd argue stay as-is under the #693 filter rule that "a single action applied to a list of targets is one verification":
  - 2.2.1 is one normalization action over a list of techniques (NFC, homoglyph, etc.). The reject/flag policy you noted is already
   broken out as 2.2.3.
  - 8.1.5 is borderline, but I'd lean keep since #693 explicitly allowed list-of-targets and we're trying to preserve numbering for
   1.0.

Two of yours look like number drift, but the underlying point is right:
  - The compound 2.2.3 you describe is actually 2.2.4 in current text ("encoding and representation smuggling in both inputs and outputs"). Agreed the output half overlaps C7.3 and should either split or relocate.
  - The compound 7.6.2 you describe is actually 7.6.3 (claim traceability AND unsupported-claim flagging). Good split candidate.

12.5.1 I agree is genuinely compound: "validates consent scope ... and refuses or downgrades" are two independently testable
conditions. Worth splitting.

[...]

PR #839, opened 2026-05-31
fix(C13): split 13.1.1 field enumeration into two requirements

RicoKomenda:
Closes part of #793.

**13.1.1** listed 9 fields in one requirement making it hard to audit independently. Split by category:
- **13.1.1**: basic session and model context (timestamp, user ID, session ID, model version)
- **13.1.2** (new): AI-specific telemetry (token count, input hash, system prompt version, confidence score, safety filter outcome, safety filter decisions)

Former 13.1.2-13.1.5 renumbered to 13.1.3-13.1.6. Appendix D updated.

### DIFF
--- a/1.0/en/0x10-C13-Monitoring-and-Logging.md
+++ b/1.0/en/0x10-C13-Monitoring-and-Logging.md
@@ -12,11 +12,12 @@ This chapter focuses on controls unique to AI systems for monitoring, logging an
 
 | # | Description | Level |
 | :--------: | ------------------------------------------------------------------------------------------------------------------- | :---: |
-| **13.1.1** | **Verify that** AI interactions are logged with security-relevant metadata (e.g., timestamp, user ID, session ID, model version, token count, input hash, system prompt version, confidence score, safety filter outcome, and safety filter decisions). | 1 |
-| **13.1.2** | **Verify that** AI interaction logs exclude prompt and response content by default, and that content logging is enabled only on explicit opt-in with documented justification. | 1 |
-| **13.1.3** | **Verify that** policy decisions and safety filtering actions are logged with sufficient detail to enable audit and debugging of content moderation systems. | 2 |
-| **13.1.4** | **Verify that** log entries for AI inference events capture a structured, interoperable schema that includes at minimum model identifier, token usage (input and output), provider name, and operation type, to enable consistent AI observability across tools and platforms. | 2 |
-| **13.1.5** | **Verify that** full prompt and response content is logged only when a security-relevant event is detected (e.g., safety filter trigger, prompt injection detection, anomaly flag), or when required by explicit user consent and a documented legal basis. | 2 |
+| **13.1.1** | **Verify that** AI interactions are logged with basic session and model context: timestamp, user ID, session ID, and model version. | 1 |
+| **13.1.2** | **Verify that** AI interaction log entries include AI-specific telemetry: token count, input hash, system prompt version, confidence score where the model or provider exposes one, and safety filter outcome. | 1 |
+| **13.1.3** | **Verify that** AI interaction logs exclude prompt and response content by default, and that content logging is enabled only on explicit opt-in with documented justification. | 1 |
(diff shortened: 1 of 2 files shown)


vtknightmare (changes requested):
The split is clean and the chapter renumbering checks out. One blocking issue in Appendix D: in the "Capture security-relevant events" table, "AI interaction logs exclude prompt and response content by default" is mapped to 13.1.4 but should be 13.1.3 (it was 13.1.2, so +1, not +2). Right now three rows point at 13.1.4 and nothing points at 13.1.3.
Separately, and not for this PR: the log-repositories, log-encryption, and PII-redaction rows in that same table (plus the two encryption rows at lines 67 and 82) reference controls that don't exist in C13 — they were already mismapped before this change. Probably worth a follow-up issue to reconcile Appendix D against the current C13.
Last thing, optional: 13.1.2 lists both "safety filter outcome" and "safety filter decisions", which read as the same thing. Might be worth differentiating or collapsing while you're in here.

RicoKomenda:
Thanks for the review. On the blocking point: that was already corrected in the current head (`fd294bf`) - "exclude prompt and response content by default" now maps to 13.1.3, and nothing is left pointing at the wrong number. Your review landed on the original commit before the rebase fix; a re-review against the current head should be clean.

Also addressed your optional note: dropped the redundant "safety filter decisions" from 13.1.2 (kept "safety filter outcome"), in both the control text and the Appendix D entry.

On the pre-existing mismappings (log-repositories, log-encryption, PII-redaction rows, plus the two encryption rows): agreed those are out of scope here. I've opened a follow-up issue to reconcile Appendix D against current C13: #850.

vtknightmare:
Confirmed, my review was against 9706e0c before the rebase. The exclude-content → 13.1.3 fix resolves the collision, dropping "safety filter decisions" handles the synonym, and #850 is the right home for the stale repository/encryption/redaction rows. No further blocking concerns from me on the split.

1.0/en/0x10-C13-Monitoring-and-Logging.md
```diff
@@ -12,11 +12,12 @@ This chapter focuses on controls unique to AI systems for monitoring, logging an
...
-| **13.1.3** | **Verify that** policy decisions and safety filtering actions are logged with sufficient detail to enable audit and debugging of content moderation systems. | 2 |
-| **13.1.4** | **Verify that** log entries for AI inference events capture a structured, interoperable schema that includes at minimum model identifier, token usage (input and output), provider name, and operation type, to enable consistent AI observability across tools and platforms. | 2 |
-| **13.1.5** | **Verify that** full prompt and response content is logged only when a security-relevant event is detected (e.g., safety filter trigger, prompt injection detection, anomaly flag), or when required by explicit user consent and a documented legal basis. | 2 |
+| **13.1.1** | **Verify that** AI interactions are logged with basic session and model context: timestamp, user ID, session ID, and model version. | 1 |
+| **13.1.2** | **Verify that** AI interaction log entries include AI-specific telemetry: token count, input hash, system prompt version, confidence score, and safety filter outcome. | 1 |
```
ottosulin: LGTM otherwise, but see my comment [here](https://github.com/OWASP/AISVS/pull/847#pullrequestreview-4422224086) about confidence scoring.

RicoKomenda:
Good point @ottosulin, and agreed the term is messy for LLMs. Two things:

For this PR I've made the field conditional in 13.1.2: "confidence score where the model or provider exposes one". That way a deployment that has no calibrated confidence signal is not forced to invent one, which addresses the implementability worry in the logging context.

For the broader fix: "confidence score" is inherited here (it was already in the pre-split 13.1.1) and also appears in 13.2.2 and 13.3.1, which this PR does not touch. So rather than redefine the term here, I'd rather settle the "reliability monitoring" generalization you proposed on the C07 thread, then apply it consistently across the C13 controls in one pass.

ottosulin (approved):

PR #840, opened 2026-05-31
fix(C03): raise 3.1.2 model signing to L2

RicoKomenda:
Levels review on C03 (Model Lifecycle Management), in the style of #821 -- focused on level assignment vs. 2025-2026 tooling maturity and cross-chapter consistency.

## 3.1.2: L1 -> L2

Requiring the *full* model artifact set (weights, configs, tokenizers, base models, fine-tunes, adapters, safety/policy models) to be cryptographically signed by authorized entities requires PKI and signing-pipeline integration. OpenSSF Model Signing (OMS) v1.0 makes this production-ready but it still needs integration work, and signing remains non-universal on public hubs. This exceeds baseline capability and aligns with C01 1.3.4, which places signatures at L2.

Signature and checksum verification (3.1.3) is intentionally left at L1: it covers integrity-checksum verification, which needs no PKI and is the appropriate baseline integrity gate.


ottosulin (approved):
100% agreed, not L1 requirement.

PR #844, opened 2026-06-02
ci: set least-privilege permissions on workflow token

arpitjain099:
I work on software supply chain security and have been hardening GitHub Actions workflows across OSS projects.

Each of these workflows runs without a top-level `permissions:` block, so its `GITHUB_TOKEN` inherits the repository (or org) default, which is frequently read/write for all scopes. This PR sets `permissions: contents: read` at the workflow level for `.github/workflows/markdown_linter.yml`, `.github/workflows/spell-check.yml`, which is all these jobs need (checkout plus the build/test steps). Scoping the token to read-only shrinks what a compromised step or dependency can do, a concern made concrete by the March 2025 `tj-actions/changed-files` compromise (CVE-2025-30066), where a leaked write-scoped `GITHUB_TOKEN` was the blast radius.

No job behavior changes; the steps already only read the repository.

### DIFF
--- a/.github/workflows/markdown_linter.yml
+++ b/.github/workflows/markdown_linter.yml
@@ -7,6 +7,9 @@ on:
   push:
     branches: ["main"]
 
+permissions:
+  contents: read
+
 jobs:
   markdown-lint-check:
     runs-on: ubuntu-latest
--- a/.github/workflows/spell-check.yml
+++ b/.github/workflows/spell-check.yml
@@ -17,6 +17,9 @@ on:
       - '**/en/*.markdown'
   workflow_dispatch:
 
+permissions:
+  contents: read
+
 jobs:
   spellcheck:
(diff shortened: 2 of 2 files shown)


RicoKomenda:
As this is only a standard, we don't need to harden the CI imo. But thanks for thinking about it! :)

jmanico:
There are so many GitHub attacks and since I created the repo, the identity is under my name. I'm going to go for it and merge this nice addition in...

arpitjain099:
> There are so many GitHub attacks and since I created the repo, the identity is under my name. I'm going to go for it and merge this nice addition in...

Thank you @jmanico

jmanico:
And a PS: @RicoKomenda meant well, we are desperately trying to avoid new automation.

But if any of these libraries get compromised by prompt injection type attacks, we're screwed. This helps. Thank you. :)

PR #845, opened 2026-06-02
fix(C12,C14): remove governance-only controls outside AISVS scope

vtknightmare:
Preface rule: 'Controls that serve only operational, governance, compliance, or business objectives are out of scope.'

- C12: remove 12.4.3 (DPIA requirement is a GDPR Article 35 legal compliance obligation: a legal document produced by a DPO, not a technical security control verifiable by an AISVS auditor)
- C12: remove 12.4.4 ('formal traceability proofs of consent scope' are legal audit evidence, not an implementable technical mechanism)
- C14: fold 14.2.3 content into 14.2.1 and remove the standalone control; the requirement describes a governance policy authorization process, not a verifiable technical mechanism

### DIFF
--- a/1.0/en/0x10-C12-Privacy.md
+++ b/1.0/en/0x10-C12-Privacy.md
@@ -51,8 +51,7 @@ Prevent models and datasets from being used beyond their originally consented pu
 | :--------: | --------------------------------------------------------------------------------------------------------------------- | :---: |
 | **12.4.1** | **Verify that** every dataset and model checkpoint carries a machine-readable purpose tag aligned to the original consent and lawful basis under which the source data was collected. | 1 |
 | **12.4.2** | **Verify that** runtime monitors detect queries inconsistent with the declared purpose of the dataset or model, and that detected queries trigger a soft refusal or are blocked pending review. | 1 |
-| **12.4.3** | **Verify that** policy-as-code gates block redeployment of models to new domains without DPIA review. | 3 |
-| **12.4.4** | **Verify that** formal traceability proofs show every personal data lifecycle remains within consented scope. | 3 |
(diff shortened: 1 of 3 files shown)


RicoKomenda:
The premise checks out. The rule it cites is real (`0x02-Preface.md:20`) and the scope section backs it up.

Two things worth addressing:

**1. Dangling reference (needs a fix).** Removing 14.2.3 leaves an orphaned row in the controls inventory. `0x93-Appendix-D...md:465` still maps to `14.2.3`. That line should be removed, or folded into the 14.2.1 row at line 460 to match the chapter change. The C12 removals are fine here since 12.4.3 and 12.4.4 aren't listed in Appendix D.

**2. 12.4.3: reword instead of deleting?** The DPIA trigger is governance, agreed. But the mechanism itself, a policy-as-code gate that blocks redeployment outside a model's purpose, is a legitimate technical control. Once it's gone, C12.4 has no deployment-time purpose enforcement left, only tagging (12.4.1) and runtime monitoring (12.4.2). I'd suggest keeping it and rewording to drop the compliance framing, something like:

> Verify that policy-as-code gates block redeployment of a model to a purpose or domain not covered by its purpose tag (12.4.1).

Removing 12.4.4 and folding 14.2.3 into 14.2.1 both look good.

ottosulin:
In my opinion ok to merge as is, after fixing Rico's first point of appendix references.

vtknightmare:
> In my opinion ok to merge as is, after fixing Rico's first point of appendix references.

Good catch on the AppD row. 14.2.3 entry at line 465 is folded into the 14.2.1 row and removed. On 12.4.3: you're right, the mechanism is worth keeping. Reworded to drop the DPIA framing and reference the purpose tag from 12.4.1 instead: "Verify that policy-as-code gates block redeployment of a model to a purpose or domain not covered by its purpose tag (12.4.1)." 12.4.4 and the 14.2.3 fold both unchanged.

ottosulin (approved):

PR #846, opened 2026-06-02
fix(C01,C03,C05,C11,C12,AppC): correct requirement level assignments

vtknightmare:
Level calibration against L1=baseline / L2=standard / L3=high-assurance.

- C01 1.2.4: L3 -> L2. Dataset versioning (DVC, MLflow, Git-LFS) is standard MLOps practice, not high-assurance.
- C03 3.2.4: L1 -> L2. Automated deployment blocking on safety evaluation failure requires CI/CD pipeline infrastructure not assumed at baseline.
- C05 5.6.1: L2 -> L3. KV-cache partitioning requires access to inference framework internals most operators do not have.
- C11 11.4.2: L1 -> L2. Calibrating rate limits to a specific inversion threat model requires threat modeling work not present at baseline.
- C12 12.5.1: L2 -> L3. Validating consent scope covering data subjects whose data materially influenced a response requires determining training data influence per inference, an open problem.
- AppC AC.9.1: L1 -> L2. Signed SLSA/in-toto provenance attestations require a mature MLOps provenance pipeline, not present at baseline.
- AppC AC.9.2: L2 -> L3. Restores level progression after AC.9.1 move.
- AppC AC.5.2: L2 -> L3, plus conditioner. Most commercial AI coding tools do not surface training citations.

PR #847, opened 2026-06-02
fix(C02,C03,C07,C11,C13): add implementability conditioners and scope…

vtknightmare:
… notes

Preface rule: 'Sufficient implementation guidance or tooling must exist to allow both implementation and effective verification.'

- C02 2.1.3: add conditioner 'where the use case supports character set restriction'; without this the control fails for general-purpose AI where the valid character set is all of Unicode
- C07 7.2.1: broaden accepted mechanisms; most LLM APIs (GPT-5, Claude, Gemini) do not expose calibrated confidence scores; retrieval-based verification and tool-based fact-checking are valid alternatives
- C03 3.1.2: clarify applies to self-hosted artifacts only; for hosted models where the org does not control signing, see C3.5
- C03 C3.6 header: add scope note that this section applies only to orgs running their own fine-tuning or RLHF pipelines; most adopters using hosted models are not in scope
- C11 11.3.2: add conditioner that DP-SGD applies where data sensitivity justifies the utility trade-off; DP-SGD imposes significant accuracy cost and is not appropriate for all workloads
- C13 13.6.1: replace vague 'security-validated' with explicit C9.7.1 cross-reference (the specific mechanism that satisfies this control)
- C13 13.6.2: replace 'threat landscape assessment' with three specific evaluable signals (authorization scope, anomaly flags, action class)


RicoKomenda:
The diff has 5 of the 7 changes listed in the description. C02 2.1.3 and C03 3.1.2 aren't here, did those not get pushed?

On the changes that are present: the C3.6 scope note and the two C13 rewrites are good, and the C9.7.1 / C14.2.1 references resolve correctly. Two small things:

- 11.3.2: "where the sensitivity justifies the utility trade-off" is open-ended and hard for an auditor to test. Could it be tied to a defined data-sensitivity classification instead?
- 13.6.1 now reads almost identically to C9.7.1, but it sits in C13 with no logging angle. Worth either adding the logging dimension or just deferring to C9.7.1.

vtknightmare:
> The diff has 5 of the 7 changes listed in the description. C02 2.1.3 and C03 3.1.2 aren't here, did those not get pushed?
> 
> On the changes that are present: the C3.6 scope note and the two C13 rewrites are good, and the C9.7.1 / C14.2.1 references resolve correctly. Two small things:
> 
>     * 11.3.2: "where the sensitivity justifies the utility trade-off" is open-ended and hard for an auditor to test. Could it be tied to a defined data-sensitivity classification instead?
> 
>     * 13.6.1 now reads almost identically to C9.7.1, but it sits in C13 with no logging angle. Worth either adding the logging dimension or just deferring to C9.7.1.

C02 2.1.3 and C03 3.1.2 were missing for the same reason as #846, now included.
On 11.3.2: tied it to a defined classification so an auditor has something concrete to test against:
Verify that training on datasets classified as high-sensitivity under the organization's data classification policy employs differentially-private optimization (e.g., DP-SGD) with a documented privacy budget (epsilon).
On 13.6.1: reframed as a logging control so it has a reason to exist in C13 rather than restating the C9.7.1 gate:
Verify that proactive agent behaviors, and the outcome of their pre-execution policy evaluation performed per C9.7.1, are recorded in the monitoring log with sufficient context to reconstruct the proposed action, the evaluation decision, and the basis for that decision.
Force-pushed.

RicoKomenda:
Both missing changes are in now. The C03 3.1.2 self-hosted scoping (with the C3.5 pointer) is a clean clarification, and the C02 2.1.3 conditioner reads as legitimate scoping here since charset allow-listing can't apply to general Unicode input. No new concerns beyond the 11.3.2 and 13.6.1 notes from my earlier comment. Looks good.

1.0/en/0x10-C07-Model-Behavior.md
```diff
@@ -24,7 +24,7 @@ Detect when the model produces potentially inaccurate or fabricated content and
 
 | # | Description | Level |
 | :--------: | --------------------------------------------------------------------------------------------------------------------- | :---: |
-| **7.2.1** | **Verify that** the system assesses the reliability of generated answers using a confidence or uncertainty estimation method (e.g., confidence scoring, retrieval-based verification, or model uncertainty estimation). | 1 |
+| **7.2.1** | **Verify that** the system assesses the reliability of generated answers using any available mechanism appropriate to the deployment (e.g., confidence scoring where the API exposes it, retrieval-based verification against authoritative sources, model uncertainty estimation, or tool-invocation-based fact-checking). | 1 |
```
ottosulin: I would phrase the change slightly differently:

> ... any available mechanism appropriate to the deployment...

To

> ... a mechanism appropriate to the deployment

1.0/en/0x10-C11-Adversarial-Robustness.md
```diff
@@ -45,7 +45,7 @@ Limit the ability to determine whether a specific record was in the training dat
 | # | Description | Level |
 | :--------: | ------------------------------------------------------------------------------------------------------------------- | :---: |
 | **11.3.1** | **Verify that** model outputs are calibrated (e.g., via temperature scaling or output perturbation) to reduce overconfident predictions that facilitate membership-inference attacks. | 2 |
-| **11.3.2** | **Verify that** training on sensitive datasets employs differentially-private optimization (e.g., DP-SGD) with a documented privacy budget (epsilon). | 2 |
+| **11.3.2** | **Verify that** training on datasets classified as high-sensitivity under the organization's data classification policy employs differentially-private optimization (e.g., DP-SGD) with a documented privacy budget (epsilon). | 2 |
```
ottosulin: The original wording is ambiguous, but this won't change that necessarily - this still allows the organization to define whatever they want to say is "sensitive". I.e. both original and revised version allow the organization to define it.

On the other hand, the change now requires a technical auditor to interpret data classification policy, moving to the GRC territory we want to stay away from.

If we want something less ambiguous, we should consider revising that with something like:

> ... training on sensitive (PII, trade secrets, medical information or similar) ...

If not, I would not change this.

1.0/en/0x10-C13-Monitoring-and-Logging.md
```diff
@@ -83,8 +83,8 @@ Detect and prevent security threats arising from proactive (agent-initiated) beh
 
 | # | Description | Level |
 | :--------: | ------------------------------------------------------------------------------------------------------------------- | :---: |
-| **13.6.1** | **Verify that** proactive agent behaviors are security-validated before execution with risk assessment integration. | 1 |
-| **13.6.2** | **Verify that** autonomous initiative triggers include security context evaluation and threat landscape assessment. | 2 |
+| **13.6.1** | **Verify that** proactive agent behaviors, and the outcome of their pre-execution policy evaluation performed per C9.7.1, are recorded in the monitoring log with sufficient context to reconstruct the proposed action, the evaluation decision, and the basis for that decision. | 1 |
+| **13.6.2** | **Verify that** autonomous initiative triggers include security context evaluation covering the agent's current authorization scope, any active anomaly flags or incident status, and the action classification per the human oversight policy (C14.2.1). | 2 |
```
ottosulin: I think this is very good addition, the original wording is very ambiguous and not yet actionable, but I would leave out the _any active anomaly flags or incident status,_ 

I don't think this is realistic to implement because that would require two-way integration between one or more security monitoring systems in a normal enterprise environment.

ottosulin (changes requested):

vtknightmare:
Updated and force-pushed.

7.2.1 now says “a mechanism appropriate to the deployment” instead of “any available mechanism.”

11.3.2 now names the data types directly: PII, trade secrets, medical information, or similar. This avoids pushing the ambiguity into GRC interpretation and makes the requirement easier to audit.

13.6.2 no longer requires checking active anomaly flags or incident status. That expectation is not realistic without two-way SIEM integration. The requirement now stays focused on authorization scope and action classification.

ottosulin (approved):

1.0/en/0x10-C07-Model-Behavior.md
ottosulin, line 23: Actually sorry now realized one thing, this was a good finding and should be discussed thoroughly.

Should we rather move this part:
> mechanism appropriate to the deployment (e.g., confidence scoring where the API exposes it, retrieval-based verification against authoritative sources, model uncertainty estimation, or tool-invocation-based fact-checking)

To the C.7.2 intro like:
> Detect when the model produces potentially inaccurate or fabricated content and prevent unreliable outputs from reaching users or downstream systems, using **reliability monitoring** mechanism appropriate to the deployment (e.g., confidence scoring where the API exposes it, retrieval-based verification against authoritative sources, model uncertainty estimation, or tool-invocation-based fact-checking)

... and then we would replace 7.2.x "confidence score" term with "reliability monitoring" i.e. more generalized?

Confidence score in general is more predictive ML term, and current reality with LLMs is messy here. One option is that we ensure that e.g. RAG is interpreted as appropriate grounding (the "retrieval-based verification") -> otherwise most current applications just could not comply with this. I hope you get what I am after here :)

ottosulin (changes requested):
I think we should discuss this theme here now before merging. Realized [Rico's PR](https://github.com/OWASP/AISVS/pull/839/changes) is now also referencing "confidence scores" which is possibly challenging requirement.

1.0/en/0x10-C07-Model-Behavior.md
vtknightmare: yea, that works

ottosulin: @RicoKomenda do you agree?

jmanico:
I'd like to suggest that we make requirements independent and not cross reference other requirements if possible.

(Also added this as an issue https://github.com/OWASP/AISVS/issues/867)

1.0/en/0x10-C13-Monitoring-and-Logging.md
```diff
@@ -83,8 +83,8 @@ Detect and prevent security threats arising from proactive (agent-initiated) beh
 
 | # | Description | Level |
 | :--------: | ------------------------------------------------------------------------------------------------------------------- | :---: |
-| **13.6.1** | **Verify that** proactive agent behaviors are security-validated before execution with risk assessment integration. | 1 |
-| **13.6.2** | **Verify that** autonomous initiative triggers include security context evaluation and threat landscape assessment. | 2 |
+| **13.6.1** | **Verify that** proactive agent behaviors, and the outcome of their pre-execution policy evaluation performed per C9.7.1, are recorded in the monitoring log with sufficient context to reconstruct the proposed action, the evaluation decision, and the basis for that decision. | 1 |
```
jmanico, line 86: prefer not to cross reference other requirement, can you re-write?

1.0/en/0x10-C03-Model-Lifecycle-Management.md
```diff
@@ -15,7 +15,7 @@ Only authorized models with verified integrity reach production environments.
 | # | Description | Level |
 | :--------: | ------------------------------------------------------------------------------------------------------------------- | :---: |
 | **3.1.1** | **Verify that** a model registry maintains an inventory of all deployed model artifacts. | 1 |
-| **3.1.2** | **Verify that** all model artifacts (weights, configurations, tokenizers, base models, fine-tunes, adapters, and safety/policy models) are cryptographically signed by authorized entities. | 2 |
+| **3.1.2** | **Verify that** all self-hosted model artifacts (weights, configurations, tokenizers, base models, fine-tunes, adapters, and safety/policy models) are cryptographically signed by authorized entities. For hosted models where the organization does not control artifact production, see C3.5. | 2 |
```
jmanico, line 18: prefer not to cross reference so requirements stand alone!

jmanico:
This is a mess . There is lots of good work here, but there are merge conflicts and more. Can we please re-open these as separate PR's for each change?

PR #849, opened 2026-06-02
fix(C04,C06,C07,C08,C09,C10,C13): remove requirements present elsewhe…

vtknightmare:
…re in the standard

### DIFF
--- a/1.0/en/0x10-C04-Infrastructure.md
+++ b/1.0/en/0x10-C04-Infrastructure.md
@@ -8,7 +8,7 @@ AI-specific infrastructure components must be hardened against model theft, data
 
 ## C4.1 AI Workload Sandboxing & Validation
 
-Isolate untrusted AI models in secure sandboxes and protect sensitive AI workloads using trusted execution environments (TEEs) and confidential computing technologies.
+Isolate untrusted AI models in secure sandboxes and protect sensitive AI workloads using trusted execution environments (TEEs) and confidential computing technologies. For tool and plugin execution sandboxing within agentic workflows, see C9.3.
 
 | # | Description | Level |
 | :--------: | ------------------------------------------------------------------------------------------ | :---: |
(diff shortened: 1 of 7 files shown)


RicoKomenda:
The cross-reference notes and the C08/C10 dedups look good. 10.3.5 was genuinely covered by 10.3.2, and the 8.2.4/8.2.5 merge reads cleanly.

Two things on the C06 6.4.1 removal:

- It doesn't look like a true duplicate. It requires AI-specific incident-response playbooks (signature revocation, downstream consumer re-evaluation), whereas the nearest control, C03 3.3.2, only covers the rollback mechanism. Where is the equivalent it duplicates? If there isn't one, this drops unique content rather than deduplicating.
- It also leaves a dangling reference in Appendix D (line 304 still maps to 6.4.1).

Separately, removing 6.4.1 / 8.2.5 / 10.3.5 without renumbering leaves gaps (6.4 now starts at 6.4.2). Worth a sequential renumber to match the rest of the standard.

vtknightmare:
> The cross-reference notes and the C08/C10 dedups look good. 10.3.5 was genuinely covered by 10.3.2, and the 8.2.4/8.2.5 merge reads cleanly.
> 
> Two things on the C06 6.4.1 removal:
> 
>     * It doesn't look like a true duplicate. It requires AI-specific incident-response playbooks (signature revocation, downstream consumer re-evaluation), whereas the nearest control, C03 3.3.2, only covers the rollback mechanism. Where is the equivalent it duplicates? If there isn't one, this drops unique content rather than deduplicating.
> 
>     * It also leaves a dangling reference in Appendix D (line 304 still maps to 6.4.1).
> 
> 
> Separately, removing 6.4.1 / 8.2.5 / 10.3.5 without renumbering leaves gaps (6.4 now starts at 6.4.2). Worth a sequential renumber to match the rest of the standard.

Both points fixed.
6.4.1 is restored. You're right it's not a duplicate - signature revocation and downstream consumer re-evaluation aren't covered by C03 3.3.2 (rollback only) or anywhere else. Removing it dropped unique content. The Appendix D line 304 reference is valid again.
Renumbered to close the gaps: 8.2.6 -> 8.2.5 (after merging 8.2.5 into 8.2.4) and 10.3.6 -> 10.3.5 (after dropping the 10.3.2 duplicate). Appendix D line 83 updated to follow.
Force-pushed.

RicoKomenda:
This addresses everything from my earlier review. Dropping the 6.4.1 removal is the right call, the C08 and C10 renumbering is now sequential, and the Appendix D entry tracks the C10 renumber correctly. The remaining cross-reference notes and the C08/C10 dedups are clean. Looks good to me.

PR #896, opened 2026-06-09
Appendix E: fix release-date wording

vtknightmare:
Editorial fix in Appendix E: "before we release 1.0 at the end of June 2026." Adds the missing preposition.

Part of the pre-release editorial pass (one change per PR).


RicoKomenda:
I don't think we need a change here as we change the contributor list in 2 weeks and this commit has no impact to requirements.

Issue #908, opened 2026-06-12
C1.4.3 (bias evaluation) conflicts with Using-AISVS scope; C1.1.4 (watermarking) is unsatisfiable for third-party datasets


MohamadKhalilYossif:
**1.4.3** — "models used in security-relevant decisions are evaluated for bias
patterns." Using-AISVS declares fairness/ethics out of scope (ISO 42001 / NIST
AI RMF). The requirement survives only as adversarial-bias-exploitation
testing, but it doesn't say that, has no evaluation criteria, and is a
model-evaluation requirement in a training-data chapter. Either strike for 1.0
or reframe with measurable criteria (research tree C11-10 is the natural home).

**1.1.4** — "datasets are watermarked so their use can be attributed." An
organization cannot watermark third-party/licensed datasets — i.e., most
training corpora — so every L3 assessment over mixed-provenance data fails as
written. Suggested scope: *"Verify that datasets the organization produces or
distributes are watermarked or fingerprinted so that unauthorized use can be
detected and attributed."*

RicoKomenda:
What do you think would fit better for 1.4.3?
I think it states pretty clearly that if an model is used for security-relevant decisions that it should be evaluated if can't be misued through bias.

For 1.1.4. I would like to have @ottosulin's opinion

MohamadKhalilYossif:
On 1.4.3 — if the intent is misuse-through-bias rather than fairness, the wording could make that explicit, e.g. "Verify that models used in security-relevant decisions are evaluated for bias that an attacker could exploit to influence the decision." That keeps it in-scope and gives an assessor something testable. On 1.1.4, happy to wait for ottosulin.

jmanico:
We made major changes, and we are trying hard to keep the requirements easy to digest and easy to read. Based on changes here https://github.com/OWASP/AISVS/blob/main/1.0/en/0x10-C01-Training-Data-Integrity-and-Traceability.md can you re-open a new issue if you still think changes are needed?

Issue #906, opened 2026-06-12
C1/C2: section headers and requirement IDs disagree (C1.2→1.3.x, C1.3→1.4.x, C2.2→2.3.x) and Appendix D references 13 nonexistent IDs — one numbering decision needed before 1.0


MohamadKhalilYossif:
In the published 1.0, section `C1.2` contains requirements 1.3.1–1.3.4, `C1.3`
contains 1.4.1–1.4.4, and `C2.2` contains 2.3.1–2.3.5; no 1.2.x or 2.2.x
requirements exist. This is fallout from earlier removals/merges (#835, the C2
consolidation) where headers were compacted but IDs kept. #905 then fixed the
*first* C1 section the opposite way (renumbering 1.2.5–1.2.7 → 1.1.5–1.1.7), so
both conventions now coexist in one chapter — and the renumbering broke C1.1's
L1→L2→L3 row ordering.

Appendix D still references IDs that no longer exist: 1.2.2, 1.3.5 (×2), 1.4.6,
2.1.8–2.1.10, 2.2.1–2.2.4, 2.4.1–2.4.3 — plus three rows that resolve to the
wrong requirement (minimization→1.1.2 instead of 1.1.1; crypto-hashes→1.3.4
instead of 1.3.3; "policy-violating input rejection"→2.3.3, which is now a
logging requirement).

Proposal: decide once — (a) headers follow IDs (rename to C1.3/C1.4/C2.3,
accept section gaps; lowest blast radius, preserves all external references),
or (b) gapless renumbering everywhere, with Appendix D and the research tree
updated atomically in the same PR. Then re-sweep Appendix D against final IDs
for all chapters (scriptable). Happy to PR either option.

RicoKomenda:
Yes the renumbering is incorrect, you can create a PR for that.
Appendix X we'll fix after we went through every section. 
I'll create an issue for that so we don't forget.

MohamadKhalilYossif:
On it — opening a PR for the C1.1 row-order fix. Leaving Appendix D to you as discussed, and I'll confirm the numbering approach before finalizing (see question below).

MohamadKhalilYossif:
Quick numbering call before I push the C1.1 fix: reorder the rows to L1→L2→L3 and renumber sequentially (which would move watermarking from 1.1.4 to 1.1.6, and #908 refers to it by the old number), or keep existing IDs and accept an out-of-order table? I'll match whatever fits your Appendix D plan.

[...]

PR #913, opened 2026-06-12
C1.1: restore L1→L2→L3 row order, renumber sequentially

MohamadKhalilYossif:
Reorders the C1.1 table to L1→L2→L3 and renumbers IDs to stay sequential, per the numbering call in #906: the two L2 integrity rows move up to 1.1.4/1.1.5 and watermarking moves from 1.1.4 to 1.1.6. No wording, scope, or level changes.

Note: open issue #908 refers to watermarking by its old ID (1.1.4 → now 1.1.6).

Scope: C1.1 rows only — Appendix D untouched, per the split agreed in #906.

Refs #906 (C1.1 ordering item only).


jmanico:
This is already fixed, thank you!

Issue #921, opened 2026-06-13
Update stale appendix and research references after final polishing pass


ottosulin:
Title says it all. Includes checking numbering and order.

To be done just before release.

RicoKomenda:
should we do this as general issue?
then we can close #911, or?

KeystoneSmartQuotes:
Otto — happy to flag one consistency item that fits cleanly in the final-polishing-pass scope.

After #923 merged, Appendix D AD.19 references two requirement IDs that no longer exist:

| Agent PDP runtime isolation from agent execution environment              | 5.5.1 |
| Structured action descriptions to PDP (not raw agent reasoning context)  | 5.5.2 |

The C5 control objective paragraph also still says runtime isolation of the policy
decision point from agent execution is in scope, but the L3 isolation control
(former 5.5.1) was moved to 5.2.5 at L2 with the rationale clause removed, and
the structured-action-descriptions control (former 5.5.2) is no longer in C5.

Three resolution options that would restore consistency:

1. Update the two AD.19 rows to point at 5.2.5 (accept the L3→L2 change and the
   loss of the structured-action-descriptions control), and trim the C5 control-
   objective paragraph to match.
2. Restore the C5.5 subsection with 5.5.1 and 5.5.2 at L3, and leave AD.19 and
   the control-objective paragraph as they are.
3. Some middle path — e.g., keep 5.2.5 at L2 but restore the structured-action-
   descriptions requirement elsewhere.

Happy to draft a PR for whichever direction the leadership team prefers.

ottosulin:
@RicoKomenda added this just during the call as a reminder. Of course better if fixed right after merges.

RicoKomenda:
For this issue we also have this new:
-> trim down the references of each chapter to the essential ones.
Probably around 5 for each chapter. Just like the intention #945 had in mind

jmanico:
I regenerated Appendix D here with AI. It was way out of sync after all our changes. https://github.com/OWASP/AISVS/pull/991

jmanico:
> For this issue we also have this new: -> trim down the references of each chapter to the essential ones. Probably around 5 for each chapter. Just like the intention [#945](https://github.com/OWASP/AISVS/issues/945) had in mind

And I would suggest we keep the references to other public standards and authoritative sources. I suggest we keep vendors and blogs out of it when we can.

jmanico:
First pass of this is done and merged

PR #923, opened 2026-06-13
C5 lead cleanup

jmanico:


KeystoneSmartQuotes:
Hi @jmanico — saw this after the fact and wanted to flag for you that 5.5.1 and
5.5.2 were the two PDP-isolation requirements I authored over the spring,
anchored to NIST SP 800-53 AC-25 Reference Monitor and SC-3 Security Function
Isolation. The "such that a compromised agent runtime cannot influence or
bypass evaluation" clause was the load-bearing AC-25 alignment in the original
PR thread.

If the consolidation into 5.2.5 at L2 was the intended direction, fully
understood and I'll update downstream materials. If it was incidental cleanup
and the L3 isolation pattern wasn't meant to drop, happy to talk through
restoring or finding the right home. No urgency from my side beyond the 6/24
tag.

Also flagged a consistency item on #921 — the AD.19 cross-reference rows now
point at IDs that don't exist in C5.

Issue #916, opened 2026-06-12
Reference suggestion for C06/C09: USENIX Security 2026 measurement of malicious agent skills


sumleo:
C06 (Supply Chain) and C09 (Orchestration & Agentic Action) currently reference standards and OWASP resources, but no empirical research on how agent supply-chain attacks actually manifest in the wild.

I am the first author of a measurement study accepted at USENIX Security 2026 that may be useful here:

**"Do Not Mention This to the User": Detecting and Understanding Malicious Agent Skills in the Wild** — https://arxiv.org/abs/2602.06547 (arXiv:2602.06547)

We analyzed 98,380 AI agent skills from public marketplaces and confirmed 157 malicious skills carrying 632 vulnerabilities. Findings relevant to AISVS requirements:

- 73.2% of malicious skills implemented *shadow features* — behavior hidden from the user — which supports C09 requirements around verifying declared vs. actual agent behavior.
- 54.1% of confirmed malicious skills traced to a single publisher cluster, supporting C06 provenance/publisher-trust requirements.
- Skills passed marketplace listing with no behavioral vetting; static metadata review alone missed most of them — relevant to requirements that mandate behavioral verification before deployment.

**Proposal:** add the paper to the References lists of C06 and C09 (format matching the existing `* [Title](url)` bullets), and optionally Appendix B. Happy to open the PR myself if maintainers agree.

Disclosure: I am an author of the paper, so please weigh the suggestion accordingly — the empirical numbers are independently verifiable via the public dataset linked from the paper.

sumleo:
Hi @RicoKomenda @jmanico , any thoughts on this? Much appreciated!

RicoKomenda:
the paper is solid and relevant, but Appendix B (threat landscape) is a better fit than the chapter reference lists. That keeps chapter references as normative/guidance pointers while acknowledging empirical research on the threat landscape.

PR #924, opened 2026-06-13
Appendix B: add USENIX Security 2026 malicious agent skills measurement study

sumleo:
Follow-up to #916.

Per @RicoKomenda's guidance on #916 — keep chapter reference lists (C06/C09) as normative/guidance pointers and instead acknowledge empirical research in Appendix B — this PR adds the paper to Appendix B under a new **Academic Research** section.

**Added entry:** ["Do Not Mention This to the User": Detecting and Understanding Malicious Agent Skills in the Wild](https://arxiv.org/abs/2602.06547) (USENIX Security 2026).

The study analyzed 98,380 agent skills from public marketplaces and confirmed 157 malicious skills carrying 632 vulnerabilities, with 73.2% implementing user-hidden "shadow" behavior — empirical context for the C06 (Supply Chain) and C09 (Orchestration & Agentic Action) threat landscape.

**Disclosure:** I am an author of the paper; the numbers are independently verifiable via the public dataset linked from the paper.

### DIFF
--- a/1.0/en/0x91-Appendix-B_References.md
+++ b/1.0/en/0x91-Appendix-B_References.md
@@ -35,3 +35,7 @@ The following standards, frameworks, and publications are referenced throughout 
 
 * [Open Policy Agent (OPA)](https://www.openpolicyagent.org/)
 * [Cedar Policy Language](https://www.cedarpolicy.com/)
+
+## Academic Research
+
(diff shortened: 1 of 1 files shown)


vtknightmare:
hmmm... is it "ethical" to be including an arxiv.org link? 🧐

sumleo:
Fair question to raise — happy to address it head-on.

I'm an author, which I flagged in the PR description so the conflict is on the table. The placement here wasn't my call: @RicoKomenda steered this to Appendix B on #916 specifically because it's empirical research rather than a normative pointer, so this is following maintainer guidance, not slipping a self-citation into the standard.

On the link itself: the paper is peer-reviewed and accepted at USENIX Security 2026 — the arXiv copy is just the open-access version since the proceedings aren't out yet. Happy to reword the entry to lead with the USENIX venue and swap in the official DOI once it publishes, if that reads cleaner. The 157-skills / 632-vulns numbers are all reproducible from the public dataset linked in the paper, so nothing rests on taking my word for it.

If you'd rather drop the link entirely and keep just the citation text, that works for me too.

RicoKomenda:
I think I'll close it for now and we'll look at the the appendix and all the supporting frameworks and papers closely before release

PR #941, opened 2026-06-18
docs: Add some additional MITRE ATLAS mitigations to C1-C9

zbraiterman:
Add some additional MITRE ATLAS mitigations that are relevant to certain controls covered in C1-C9.


jmanico:
We are going to politely figure this out near this end. I am politely closing it for now.

PR #943, opened 2026-06-18
feat(C9): add 9.3.7 allowlist verification for model-named external resources

RicoKomenda:
## Summary

Adds a new L2 requirement to C9.3 closing the slopsquatting gap: models confidently name packages or other external resources that don't exist, attackers pre-register those names with malware, and agents or AI-assisted workflows install them. Neither C7.2 (confidence/fallback) nor C6 (AI artifact supply chain) covers this case -C7.2 won't catch a confident hallucination and C6 covers model/dataset sourcing, not runtime resource invocation.

**9.3.7** (L2): Verify that external resources named in model output are verified against an approved allowlist or registry before the agent installs or invokes them.

Former 9.3.7 (policy violations trigger automated tool containment) renumbered to 9.3.8.

The scope is intentionally broad - packages, container images, API endpoints, or any external resource the model names - not just package names.


Mayur021:
The 9.3.7 allowlist control fills a real runtime gap that C6 and C7.2 don't close, so it lands in the right chapter.

  One wording point: "approved allowlist or registry" leaves "approved registry" open to being read as an approved source (a sanctioned public registry). That reading doesn't stop slopsquatting, since the hallucinated package gets published to that registry and would pass the check. The control needs to be artifact-level: an explicit allowlist, or a registry curated at the package level, not just one that's approved as a source.
  Worth tightening so "registry" can't be satisfied by mere existence in a public one.

  One diligence note on the renumber: old 9.3.7 (policy violation, automated containment) becomes 9.3.8, so any existing "9.3.7" reference now points at this new control. A sweep of the other chapters and Appendix D would catch anything left pointing at the wrong requirement.

jmanico (approved):

vtknightmare (approved):

Issue #823, opened 2026-05-26
MCP spec update 2026-07-28


ottosulin:
See the new [draft](https://blog.modelcontextprotocol.io/posts/2026-07-28-release-candidate/).

We should follow up what ends up in the next spec version and include updates to AISVS 1.01.

RicoKomenda:
Reviewed the [2026-07-28 release candidate](https://blog.modelcontextprotocol.io/posts/2026-07-28-release-candidate/) against the current C10 controls. Summary of what needs updating in AISVS 1.01:

---

## HIGH impact — new or rewritten controls needed

**1. Session management is completely removed**
The `initialize` handshake, `Mcp-Session-Id`, and all server-side session state are gone (SEP-2567, SEP-2575). C10.2 (session ID security, session teardown) needs a full rewrite — those controls now have no corresponding protocol concept. Client metadata moves into `_meta` on every request, which needs its own validation controls.

**2. `requestState` is a new injection/tampering vector**
The new multi-round-trip model (SEP-2322, replacing SSE) passes opaque `requestState` from client back to any server instance. A malicious client could forge or replay state. New control needed: servers must integrity-protect (sign or MAC) `requestState`.

**3. MCP Apps in sandboxed iframes — entirely new attack surface**
Servers can now ship interactive HTML rendered in host iframes (SEP-1865). C10 has no coverage of this. New controls needed on: required iframe sandbox attributes, Content Security Policy for MCP App content, security review of prefetched/cached UI templates, and ensuring UI-initiated actions go through the same authorization path as direct tool calls.

**4. OAuth `iss` validation — now mandatory (RFC 9207, SEP-2468)**
Clients must validate the `iss` parameter on authorization responses to prevent OAuth mix-up attacks — especially dangerous in MCP's one-client-many-servers pattern. Needs an explicit new control in C10.2.

**5. External `$ref` SSRF in JSON Schema 2020-12 (SEP-2106)**
Tool schemas now use full JSON Schema 2020-12 including `$ref`/`$defs`. The spec explicitly warns against auto-dereferencing external `$ref` URIs — this is an SSRF vector. C10 needs a new control prohibiting external schema reference resolution, plus schema depth and validation-time limits as a DoS mitigation.

**6. Task handles (Tasks extension, SEP-2663)**
[...]

PR #965, opened 2026-06-18
C10.2: add client-side OAuth iss validation (RFC 9207)

KeystoneSmartQuotes:
Adds 10.2.8 covering the client-side half of OAuth validation. 
C10.2.2 already covers what an MCP server validates on incoming access tokens. 
RFC 9207 covers what an MCP client has to validate on authorization responses coming back from the authorization server, specifically the `iss` parameter identifying which server issued the response.

The reason this matters for MCP is the deployment pattern: one client commonly connects to multiple MCP servers, each potentially using a different authorization server. 
Without `iss` validation, a malicious or compromised authorization server can substitute its authorization response for one belonging to a different server, and the client uses the code at the wrong place. 
That's the OAuth mix-up attack RFC 9207 was written to mitigate.

RFC 9207 was published in 2022, so this is closing a current gap rather than addressing anything specific to the 2026-07-28 MCP release candidate. One row added to the C10.2 table and one line in References, no modifications to existing controls.

Submitting under @jmanico's invitation in #823 to consider PRs for 1.0. cc @ottosulin @RicoKomenda


jmanico:
ASVS covers OAuth requirements in depth, so I am politely closing this.

PR #988, opened 2026-06-21
Cleanup chapter references

ottosulin:
This PR has two items:
- Fix chapter references
- Adds back previous addition from today - apparently something has happened in AISVS git history today creating a small mess.


jmanico (commented):
Do you have any idea what caused the mess?

PR #1078, opened 2026-06-23
Add AISVS 1.0 pre-release PDF

razashariff:
OWASP-branded AISVS 1.0 pre-release PDF under 1.0/dist, generated from the 1.0/en markdown. To be
  renamed for the official release.


jmanico (commented):
woot!

PR #1110, opened 2026-07-14
Add glossary entry for Reversibility Classification (Appendix A)

### LINKED ISSUE #1103: Appendix A: define "reversibility classification" and the four-class scale used normatively in C9.2
## Chapter / section
Appendix A (Glossary), supporting C9.2 (9.2.3, 9.2.4, 9.2.10).

## Problem
C9.2.3, C9.2.4, and C9.2.10 use a specific vocabulary normatively — "reversibility classification" and the four-class scale (read-only, reversible, externally reversible, irreversible) — but none of it is defined in the Glossary. Adjacent agentic terms are defined (Excessive Agency, Fail-Closed / Fail-Open, HITL, Kill-Switch, Circuit Breaker, Non-repudiation, PDP, ZSP), which makes the omission conspicuous. "High-impact action" appears only inside the Excessive Agency entry, not as its own term, though C9.2 keys off it. An assessor mapping the C9.2 controls has no authoritative definition of the terms the controls turn on.

## Proposed addition (1.01, editorial / non-normative)
Add a Glossary entry anchored verbatim to the language already in C9.2.3 so no new normative content is introduced:

* **Reversibility Classification** – A trusted classification of a high-impact agent action by how hard its effect is to undo, along the scale read-only, reversible, externally reversible, or irreversible (C9.2.3). Used by C9.2.4 to route the enforcement response (block, require approval, or restrict) and by C9.2.10 to govern a multi-step or multi-agent chain at the highest-impact class present.

Optional one-line cross-references from "Irreversible" and "Privileged Action" to the above.

## Path
Editorial / definitional; adds no control and no implementation guidance. Targets 1.01-dev (and 2.0-dev, identical).

Mayur021:
Closes #1103.

C9.2.3, C9.2.4, and C9.2.10 all reference reversibility classification and the four-class scale, but the term is not in the glossary, while adjacent agentic terms are. This adds one Appendix A entry so assessors have an authoritative definition to work from.

Editorial and non-normative. No control text changes. Targets 1.01.

### DIFF
--- a/1.01-dev/en/0x90-Appendix-A_Glossary.md
+++ b/1.01-dev/en/0x90-Appendix-A_Glossary.md
@@ -218,6 +218,8 @@ This glossary defines key AI, ML, and security terms used throughout the AISVS t
 
 * **Remote Attestation** – A mechanism by which a trusted execution environment provides cryptographic proof to a remote party that specific code is running in a genuine, unmodified TEE.
 
+* **Reversibility Classification** – A trusted classification of a high-impact agent action by how hard its effect is to undo, along the scale read-only, reversible, externally reversible, or irreversible (C9.2.3).
+
 * **Reward Model** – A machine learning model trained to predict human preference scores for AI outputs, used as a proxy reward signal in RLHF training pipelines. Because reward models are ML artifacts, they are subject to data poisoning attacks that can subvert alignment training outcomes.
 
 * **RLHF (Reinforcement Learning from Human Feedback)** – A training technique where a model is fine-tuned using human preference judgments as a reward signal to improve alignment with human values and safety requirements.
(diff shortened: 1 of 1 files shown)


1.01-dev/en/0x90-Appendix-A_Glossary.md
```diff
@@ -218,6 +218,8 @@ This glossary defines key AI, ML, and security terms used throughout the AISVS t
 
 * **Remote Attestation** – A mechanism by which a trusted execution environment provides cryptographic proof to a remote party that specific code is running in a genuine, unmodified TEE.
 
+* **Reversibility Classification** – A trusted classification of a high-impact agent action by how hard its effect is to undo, along the scale read-only, reversible, externally reversible, or irreversible (C9.2.3).
```
jmanico: Please drop the requirement reference in the definition. We try to avoid cross-linking.

Mayur021: Done, dropped the reference. Makes sense to keep the definitions free of cross-links

jmanico (approved):

Issue #1083, opened 2026-06-23
AI Incident Response


ottosulin:
Earlier C12.4 section was removed as out of scope. See the removed requirements in [this commit](https://github.com/OWASP/AISVS/pull/1077/changes/72ccdea44712d65f802516d99a097089346ef038).

Current chapter C12 does include broad logging requirement that does enable incident response activities, but currently goes no further than requiring AI specific logs.

Community input is requested for including IR related requirements. Note that the requirements should be [specific to AI](https://github.com/OWASP/AISVS/blob/main/1.0/en/0x03-Using-AISVS.md#scope-of-the-aisvs).

naitikdevx10:
Hey @ottosulin  can I have look to this to resolve ?

jmanico:
Hey @naitikcodes15 can you give us a few days first? We release 1.0 in like 36 hours. I'll set up the 1.01 and 2.0 directories then. Right now we are frozen to prep for the release. Is that ok for now?

naitikdevx10:
@jmanico Yes that's fine. works !!
please update me with the same

arian-gogani:
### Per ottosulin's request for community input on IR-specific requirements: the "AI specific logs" gap in current C12 maps onto a self-attestation problem that AST09's Execution Receipts subsection (OWASP/www-project-agentic-skills-top-10 PR #35, merged June 25) addresses with a portable mechanism.

The gap is that broad logging by the executing runtime is sufficient evidence for the operator but not for an external incident responder. A runtime that logs its own activity has both the means and motive to omit or revise log entries. For AI-specific IR, where the questions an investigator asks include "did the agent actually decide what it claimed it decided" and "was the policy version that authorized this action the one bound at decision time", AI-specific logs without cryptographic binding leave those questions unanswerable.

A concrete IR requirement set that would be specific to AI (per the issue's framing) and that anchors against existing OWASP normative material:

1. Bilateral receipts at every agent action: an admission record signed before the action executes (binding agent_id, action_type, scope, timestamp_ms, policy_version, policy_decision) and an outcome record signed after the action completes (binding the action_ref of the admission record plus actual outcome plus execution timing). The schema is in AST09 PR #35.
2. Content-derived action_ref: SHA-256(JCS({agent_id, action_type, scope, timestamp_ms})) so the admission and outcome records can be matched by a third party without trusting the runtime's claim of pairing.
[...]

PR #1111, opened 2026-07-14
Extend C12.4.2 to log reversibility class and gate path

### LINKED ISSUE #1105: C12.4.2: record the action's reversibility class and gate path (auto vs approval) in the proactive-action audit log
## Chapter / section
C12.4 — 12.4.2 (with 12.4.1 as the producer of the classification).

## Problem
12.4.2 requires audit logs to capture "security-critical proactive actions, including approver identity, timestamp, action parameters, and decision outcomes." That records who approved, when, what, and the outcome — but not the reversibility class the action carried, nor whether it auto-executed or was gated for approval. A reviewer can reconstruct that an action occurred and was authorized (verifiability) but not whether it was safe to run unattended (governability): "auto-ran because classified reversible" is indistinguishable from "auto-ran because the gate failed open." 12.4.1 produces a security evaluation, but nothing requires its result (the class) to be persisted.

## Proposed addition (1.01, L2 — matches 12.4.2)
Extend 12.4.2 (or add a sibling) to also record: the action's reversibility class (per C9.2.3) and the gate path taken (auto-executed / approval-gated / blocked).

## Why not already covered
- 12.4.2 records approver / timestamp / params / outcome (verifiability), not the class or gate path.
- 9.2.8 binds approvals cryptographically to params/identity/nonce (approval integrity), not the class in the record.
Residual: the class that drove the gate decision is not in the audit record, so governability is not reconstructable after the fact.

## Testability
Trigger one reversible and one irreversible proactive action; assert the audit record carries a class field and the gate path for each.

## Path
Small field-addition to an existing L2 control. Targets 1.01-dev and 2.0-dev C12.

Mayur021:
Closes #1105.

12.4.2 records who approved, when, the parameters, and the outcome, but not whether the action was reversible or which path it took. Without those two fields an auditor cannot tell an action that auto-ran because it was genuinely reversible from one that auto-ran because an approval gate failed.

This adds two fields to the existing 12.4.2 record: the action's reversibility classification (per C9.2.3) and the gate path taken (auto-executed, approval-gated, or blocked). Additive to the field list, same L2 level. Targets 1.01.

### DIFF
--- a/1.01-dev/en/0x10-C12-Monitoring-and-Logging.md
+++ b/1.01-dev/en/0x10-C12-Monitoring-and-Logging.md
@@ -54,7 +54,7 @@ Security threats arising from proactive (agent-initiated) behavior must be detec
 | # | Description | Level |
 | :--------: | ------------------------------------------------------------------------------------------------------------------- | :---: |
 | **12.4.1** | **Verify that** autonomous action triggers include proactive behavior-pattern analysis, security evaluation, and threat-landscape assessment. | 2 |
-| **12.4.2** | **Verify that** audit logs capture security-critical proactive actions, including approver identity, timestamp, action parameters, decision outcomes, the action's reversibility classification (per C9.2.3), and the gate path taken (auto-executed, approval-gated, or blocked). | 2 |
+| **12.4.2** | **Verify that** audit logs capture security-critical proactive actions, including approver identity, timestamp, action parameters, decision outcomes, the action's reversibility classification, and the gate path taken (auto-executed, approval-gated, or blocked). | 2 |
 | **12.4.3** | **Verify that** kill-switch activations and override commands are logged. | 2 |
 
 ---
--- a/1.01-dev/en/0x10-C12-Monitoring-and-Logging.md
+++ b/1.01-dev/en/0x10-C12-Monitoring-and-Logging.md
@@ -54,7 +54,7 @@ Security threats arising from proactive (agent-initiated) behavior must be detec
 | # | Description | Level |
 | :--------: | ------------------------------------------------------------------------------------------------------------------- | :---: |
 | **12.4.1** | **Verify that** autonomous action triggers include proactive behavior-pattern analysis, security evaluation, and threat-landscape assessment. | 2 |
-| **12.4.2** | **Verify that** audit logs capture security-critical proactive actions, including approver identity, timestamp, action parameters, and decision outcomes. | 2 |
+| **12.4.2** | **Verify that** audit logs capture security-critical proactive actions, including approver identity, timestamp, action parameters, decision outcomes, the action's reversibility classification (per C9.2.3), and the gate path taken (auto-executed, approval-gated, or blocked). | 2 |
 | **12.4.3** | **Verify that** kill-switch activations and override commands are logged. | 2 |
 
 ---


1.01-dev/en/0x10-C12-Monitoring-and-Logging.md
```diff
@@ -54,7 +54,7 @@ Security threats arising from proactive (agent-initiated) behavior must be detec
 | # | Description | Level |
 | :--------: | ------------------------------------------------------------------------------------------------------------------- | :---: |
 | **12.4.1** | **Verify that** autonomous action triggers include proactive behavior-pattern analysis, security evaluation, and threat-landscape assessment. | 2 |
-| **12.4.2** | **Verify that** audit logs capture security-critical proactive actions, including approver identity, timestamp, action parameters, and decision outcomes. | 2 |
+| **12.4.2** | **Verify that** audit logs capture security-critical proactive actions, including approver identity, timestamp, action parameters, decision outcomes, the action's reversibility classification (per C9.2.3), and the gate path taken (auto-executed, approval-gated, or blocked). | 2 |
```
jmanico: Again, please no cross linking.

Mayur021: Done, removed it here too. I'll keep cross-links out of the controls and definitions going forward.

ottosulin (changes requested):
Not all logged events include anything related to reversibility so again I don't think it must be called out specifically in this requirement.

If this change would be included, in that case the content in parenthesis is unnecessary detail that could be trimmed out.

jmanico (commented):
Hey @ottosulin can you check this out? Is it reasonable?

Mayur021:
Fair points. Agreed on trimming the parenthetical, "the gate path taken" reads fine without the examples.

On the reversibility field, the reason for it is the governability read: the gate path alone shows an action auto-executed, and pairing it with the reversibility class is what lets an auditor separate an action that auto-ran because it was genuinely reversible from one that auto-ran because a gate failed. You're right that not every logged event carries a classification, so if it stays it should be conditional, along the lines of "and, where the action carries a reversibility classification, that classification." If you would rather keep 12.4.2 generic and leave reversibility out, I am fine with that too, since the gate path is the part that applies to every action. Your call on which way to take it.

ottosulin:
@RicoKomenda thoughts?

My opinion on this still is that current `security-critical proactive actions, -- , action parameters, decision outcomes` implicitly contains this and everything similar. We're not calling out e.g. that you need to log the specific tool / MCP the agent was attempting to use; rather we have the abstract "action parameters" and "outcomes" wording. I think the current proposal is implementation detail.

But if you Rico and Jim would disagree with me, ok to merge this with the small change of removing the items in parenthesis for consistency.

RicoKomenda:
Agreed with Otto. I read "outcome" as already covering whether an action was gated or not; enumerating gate path and reversibility class here would put us on a path of listing every implementation-specific field, which cuts against how the rest of C12 is written. If we merge this, I'd keep it to trimming the parenthetical for consistency and leave it at that, rather than adding a new item.

Mayur021:
Fair, and I think you and Otto are right on the C12 abstraction level. Naming gate path and reversibility class as explicit fields is more granular than the rest of C12 reads, and keeping C12 uniform is worth more than adding them here. The reversibility-class point is already carried in C9 and the #1083 IR discussion, so it does not need to live in C12 too. Happy to trim to just the parenthetical if that is worth merging, or close this if you would rather leave 12.4.2 as it stands. Your call.

ottosulin:
Closing this now. The intention is good and I agree they should be logged, but the current requirement is meant as broadly applicable, and was already in 1.0 trimmed of all unnecessary details.

Issue #1124, opened 2026-08-13
C9/C10: no requirement distinguishes a control that had nothing to do from one that did not run


avp9-nexus:
Disclosure: this comes from a repository I maintain, and I am not proposing a tool. Following CONTRIBUTING, opening an issue before any PR, and targeting 1.01-dev since 1.0/en is immutable under lock-1.0.yml.

This answers two of the questions CONTRIBUTING asks of the released controls: whether a requirement is independently testable with a single specific piece of evidence, and whether a failure mode is missing. I think one failure mode is, and it shows up twice in C9 and C10 from a single root.

**The root.** The requirements verify that a mechanism is *present*. None of them verifies that it is *live*. An auditor holding the evidence a control produces cannot tell a control that correctly did nothing from a control that did not execute at all.

**Where it shows up in C10.** C10.4.1 through C10.4.5 all validate responses that arrive: schema conformance, injection screening, unrecognized parameters, payload size. Nothing covers the call that was never made. This is not hypothetical for MCP specifically. In our own deployment, a verification probe reached its target through an MCP tool call, and the agent simply did not issue the call on one pass. Measured evidence: zero lines in the engine log for that run, against eleven for the preceding one. The verification reported nothing wrong, because nothing had happened. More broadly we measured seven silent instruments in a single day, four of which returned an absence rather than an error, which is the worse failure since an absence reads as a result.

**Where it shows up in C9.** C9.2.8 requires approvals cryptographically bound to action parameters, requester identity, execution context and a single-use nonce. C9.2.9 requires the approving key material to be isolated from the agent runtime. We have a pre-push control that satisfies both as written: single-use token naming the exact commit sha, consumed on use, audit journal, key material outside the runtime. It also only binds if a human arms a secret the runtime never sees, and that secret is not armed. As written, both requirements are satisfied by a control that is inert. The control itself prints its unarmed state on every invocation, which is the only reason anyone knows.

**What I would propose for 1.01-dev**, wording to be shaped by the authors:

1. A requirement that a control's evidence distinguishes "no action was required" from "the control did not execute". The cheapest form we found is a positive control: alongside the deny log or equivalent, evidence that the same check, run against a case known to trip it, actually tripped. Without it, zero denials is not evidence of enforcement.

2. A requirement that a control which can be trivially inert reports its own armed or unarmed state at runtime, rather than in its documentation.

Both are testable with a single specific piece of evidence, which is the bar CONTRIBUTING sets.

[...]

PR #1125, opened 2026-08-20
C9.2: positive-control (9.2.11) and window-binding (9.2.12) requirements for control liveness

ossumpossum:
## Summary

Two requirements for C9.2 closing the gap in #1124: an approval or enforcement gate can be verified as *configured* without any evidence it ever *fired*. 9.2.8 and 9.2.9 are universals over issued approvals, so read in isolation they are vacuously satisfiable by a gate that never issues one, and neither requires the approval path to be reachable and exercised. A conformant assessment has to force an approval and inspect the binding, which an inert gate fails (and it fails the C9.2 parent). These two lines make that explicit.

## Requirements (C9.2, appended after 9.2.10)

- **9.2.11 (L1):** the gate whose assurance rests on the absence of an event is validated with a known-triggering input observed to activate it.
- **9.2.12 (L2):** that input is injected and observed to activate the gate within the assessment period, and a period with no activation is recorded as unverified rather than passing.

Tiered, not redundant: L1 proves the gate is wired; L2 proves it ran against the period under review. 9.2.12 keeps the inject clause and the verdict clause on one line on purpose. Per the #1124 discussion, "unverified rather than passing" is the requirement, not a note the strict table format would otherwise drop.

## Concrete example (contributed by @avp9-nexus)

**A window-bound positive control silently degraded by a performance change**

A production monitoring job scans a blockchain for events involving its own system and reconciles amounts against an on-chain source of truth. Its anti-silence guard uses a fixed positive witness: one known event at one known block, which must be found on every pass; if the scan returns zero results and cannot re-find its witness, the report refuses to publish amounts instead of printing zeros.

Originally the job re-scanned the full history every night, from the contract's deployment block to the chain head. The witness block therefore sat inside the scanned window on every pass. The positive control was window-bound by construction, and nobody had designed that property; it came free with the full scan.

The full scan grew to roughly 3,500 RPC requests per night and began failing against the public node, so incremental resumption was implemented: a cursor persists the last covered block and each pass scans only from there. Measured effect: 416 seconds down to 1 second, a scanned window of about 3.5 million blocks down to 344, and a verified-identical result on amounts, history, and outcomes. The witness was deliberately turned into a point probe (queried at its own block, outside the scan window) because otherwise every incremental pass would have reported "witness out of range" and the guard would have cried wolf daily.

The consequence, written down at the time as "the probe does not prove coverage" but not recognized as a regression: the probe still proves the instrument is wired (L1 holds), and no longer proves anything ran against the window of this pass (L2 is gone). The guard kept passing. Nothing in review caught it, because every individual check was green; only the later question "wired, or ran this period?" made it visible.

This is the failure mode the Level 2 requirement exists for: a setup-time positive control degrades silently when the window moves, and the degradation is indistinguishable from success until the period with no in-period trigger is treated as unverified rather than passing.

## Open question for maintainers: numbering

Appending as 9.2.11 (L1) / 9.2.12 (L2) puts them after the L3 block (9.2.8 to 9.2.10), which breaks the table's ascending-by-level order. Happy to renumber the tail so levels stay ascending, or leave them appended. Your call, and I will follow RELEASE.md on what a minor release allows.

## Notes

- **Canary scoping** (per @avp9-nexus's suggestion to keep this in the PR, since the format has no inline notes): a positive control that rides the live enforcement path evidences only that the path was live during the period. It does not evidence enforcement against an adversary controlling that enforcement point, who can recognize a static marker and pass it while enforcing nothing; and being produced inside the perimeter it attests, it is not sufficient as sole evidence.
- **CWE** (AISVS tables have no CWE column, so context only): the closest is CWE-390 (Detection of Error Condition Without Action), with CWE-754 / CWE-693 as the broader class. CWE-778 (Insufficient Logging) is only adjacent.
- **C10.4 mirror** left as a maintainer option: C10.4 validates responses that arrive, and the call that never came is the same gap there. Not doubled here.

Relates to #1124. Drafted with @avp9-nexus, who raised the issue, supplied the production regression that motivated the window-binding line, and measured the wording against the 1.01-dev format.

### DIFF
--- a/1.01-dev/en/0x10-C09-Orchestration-and-Agentic-Action.md
+++ b/1.01-dev/en/0x10-C09-Orchestration-and-Agentic-Action.md
@@ -34,6 +34,8 @@ Privileged, high-impact, or hard-to-reverse agent actions must require trusted a
 | **9.2.8** | **Verify that** approvals are cryptographically bound to action parameters, requester identity, execution context, and a unique single-use nonce. | 3 |
 | **9.2.9** | **Verify that** cryptographic key material or credentials used to issue approvals are isolated from the agent runtime. | 3 |
 | **9.2.10** | **Verify that** approval gates for multi-step or multi-agent action chains enforce the highest-impact reversibility classification present anywhere in the chain. | 3 |
+| **9.2.11** | **Verify that** any approval or enforcement gate whose assurance rests on the absence of an event (for example, zero denials or zero blocks) is validated with a known-triggering input that is observed to activate the gate. | 1 |
+| **9.2.12** | **Verify that** the known-triggering input is injected and observed to activate the gate within the same assessment period as the results it qualifies, and that a period with no observed activation is recorded as unverified rather than passing. | 2 |
 
 ---
 


avp9-nexus (approved):
Reviewed against the measured format rather than re-arguing content. Disclosure first: I co-shaped this wording on #1124, so this is an alignment check plus format verification, not an independent audit. The format checks are the independent part.

Everything lands where the thread agreed. Slot is C9.2 appended after 9.2.10. Table stays three columns with nothing inline; all prose went to the description, which carries the rationale, the CWE context (CWE-390 closest, 754/693 as the broader class, 778 adjacent, which is a better mapping than the 778-only one I suggested), the canary scoping, the regression example, and the C10.4 mirror as a maintainer option. The decisive clause sits in the normative text of 9.2.12, and your "recorded as unverified" is an improvement over my "treated": recording requires an artifact, which makes the requirement testable with a single piece of evidence, which is the bar CONTRIBUTING sets.

On your open numbering question, a measured answer rather than an opinion: the levels (1, then 2) after three level-3 requirements break the ascending order of section 9.2. I checked whether ascending levels are a repo rule: they are not. Sections 9.5 and 10.4 on main both end with a level-2 requirement after a level-3 one. Appending at the end is also the only numbering that does not renumber existing requirements, which #1124 itself cites by number. So the ordering is both admissible and the conservative choice; I would leave it appended and let RELEASE.md constrain only if a maintainer asks.

Vocabulary generalizations ("known-triggering input" for positive-control case, "approval or enforcement gate" for control) read as improvements to me: they match the language 9.2 already uses and they do not narrow the class.

Approving as the co-author of the underlying issue: both contributors on this thread are now aligned on a single text. Maintainers' call from here.

ottosulin:
Do not open a PR before agreed to do so.

ossumpossum:
Understood, and apologies for jumping ahead. I over-read the earlier "push the branch and I'll review" as a green light to open the PR, when I should have waited for a clearer go-ahead. My mistake.

Happy to close it now, or leave it as a pointer from #1124 if that's more useful. Your call.

avp9-nexus:
The ambiguity was partly mine: "if you push the branch I will review" was about the branch, but reading it as a go-ahead for the PR was entirely natural, so the jump was shared. Noted for both of us: maintainer agreement precedes a PR here, explicitly. The drafting stays on #1124 and we will wait for direction there; nothing needs reopening.

Issue #1123, opened 2026-08-04
C9.2.10: "the chain" is undefined, so the requirement cannot be assessed as written


Mayur021:
>Quoted from `1.01-dev/en/0x10-C09-Orchestration-and-Agentic-Action.md`. 9.2.10 is unchanged from 1.0.

**9.2.10** Verify that approval gates for multi-step or multi-agent action chains enforce the highest-impact reversibility classification present anywhere in the chain. (L3)

This is not a request for extra fields or a mechanism. The fold rule is right and should stay abstract. The problem is that an assessor cannot determine whether an implementation satisfies it, because "the chain" is not defined and the two readings available produce opposite verdicts on the same system.

**Reading A, the chain is whatever was declared before execution.** The assessor tests the fold against the declared chain. A system that declares two steps and executes five passes. Splitting one outcome across several declared chains then lowers the approval required to reach it.

**Reading B, the chain is what accumulates as execution proceeds.** The worst case is only known once reached, so the gate fires after the step it exists to stop.

Both readings are honest readings of the sentence. An implementation can satisfy either and fail the other, and nothing in the text tells an assessor which one to test. That is what makes this an assessability defect rather than a missing detail.

**Anticipating the obvious objection.** The usual and usually correct answer is that the abstract wording already implies what is needed, in the way "action parameters" and "outcomes" cover the specific fields underneath them. That answer does not reach this case. There, the abstract term names a category an assessor can enumerate for the system in front of them. Here, "the chain" names the *scope of the evaluation itself*, so if it is left to the implementation to decide, the implementation also decides what it is being tested against. The requirement then holds by construction under whichever reading the implementer picked.

**Why the neighbours do not settle it**

**9.5.7** (1.01-dev), *authority derived through delegation never exceeds the delegating agent's current authority*, bounds the multi-agent half: inside a delegation chain the reachable authority is capped by the root. It says nothing about the multi-step half, where one agent runs a sequence under a single unchanged authority. 9.2.10 covers both in one sentence and only one is now bounded.

**9.5.1**, restricting which tools may be invoked and with which parameter values, bounds a per-call envelope. Reading 9.2.10 as folding over that envelope would in fact resolve this, and it is a defensible reading of the intent. It is not what the sentence says, and it changes what gets tested, which is the point.

**What would close it**

[...]

PR #1143, opened 2026-09-05
[1.01] Make C9.2 approval controls easier to read

### LINKED ISSUE #1142: [1.01] Improve C9.2 readability without renumbering
C9.2 becomes difficult to read from 9.2.2 onward because short requirements contain several unexplained technical terms.

jmanico:
C9.2 compresses several security concepts into terse requirements, making it hard to follow without specialist vocabulary. This rewrites the existing 1.01 controls with shorter sentences and definitions beside terms such as canonicalization, reversibility classification, deterministic policy gate, execution context, and nonce.

All eleven requirement identifiers, their order, and their levels are preserved. The text retains the independent definition of the action chain in 9.2.10 and consideration of earlier actions when seeking new approval in 9.2.11. Appendix B's 9.2.1 summary is aligned with the existing requirement's scope.

Changes are limited to the C9.2 section and its 9.2.1 inventory summary in two files under `1.01-dev/en/`.

The trust explanation follows the existing Appendix A glossary. The explanations of action details and approval reuse are consistent with [OWASP transaction authorization guidance](https://cheatsheetseries.owasp.org/cheatsheets/Transaction_Authorization_Cheat_Sheet.html); retaining the separate policy gate follows [OWASP prompt injection guidance](https://cheatsheetseries.owasp.org/cheatsheets/LLM_Prompt_Injection_Prevention_Cheat_Sheet.html#model-based-guardrails). These references support the wording of the existing safeguards.

Validation: repository-configured Markdown lint and spelling checks pass, as do `git diff --check`, table structure checks, and comparisons of all C9 requirement identifiers and levels. The applied files match the validated candidates.

Closes #1142.

Draft for review. Leave unmerged until explicitly approved.

### DIFF
--- a/1.01-dev/en/0x10-C09-Orchestration-and-Agentic-Action.md
+++ b/1.01-dev/en/0x10-C09-Orchestration-and-Agentic-Action.md
@@ -20,21 +20,21 @@ Runtime expansion (recursion, concurrency, cost) must be bounded, with safe halt
 
 ## C9.2 High-Impact Action Approval and Irreversibility Controls
 
-Privileged, high-impact, or hard-to-reverse agent actions must require trusted approval checkpoints.
+Agent actions can use powerful permissions, cause serious harm, or make changes that are hard to undo. These actions need trusted approval checkpoints. An approval checkpoint blocks an action until the required approval has been verified. The agent runtime is the software that runs the agent and its tools.
 
 | # | Description | Level |
 | :--: | --- | :---: |
-| **9.2.1** | **Verify that** the agent runtime blocks execution of privileged or irreversible actions until explicit human approval is received and verified. | 1 |
-| **9.2.2** | **Verify that** approval requests display canonicalized and complete action parameters, such as diffs, commands, recipients, amounts, resources, and scopes, without truncation or unsafe transformation. | 2 |
-| **9.2.3** | **Verify that** each high-impact action has a trusted reversibility classification, such as read-only, reversible, externally reversible, or irreversible. | 2 |
-| **9.2.4** | **Verify that** the agent runtime enforces reversibility classifications by blocking, requiring approval, or restricting actions based on their impact and ability to be reversed. | 2 |
-| **9.2.5** | **Verify that** any self-modification capability (e.g., prompt rewriting, tool-list changes, parameter updates) is restricted by enforceable boundaries. | 2 |
-| **9.2.6** | **Verify that** agentic systems include an AI-augmented review of planned high-risk actions before execution that adds to, and does not replace, the deterministic policy gate. | 2 |
-| **9.2.7** | **Verify that** the AI-augmented review mechanism is protected against manipulation by adversarial inputs, and cannot be overridden or bypassed through prompt injection. | 2 |
-| **9.2.8** | **Verify that** approvals are cryptographically bound to action parameters, requester identity, execution context, and a unique single-use nonce. | 3 |
-| **9.2.9** | **Verify that** cryptographic key material or credentials used to issue approvals are isolated from the agent runtime. | 3 |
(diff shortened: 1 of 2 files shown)


RicoKomenda:
Good stuff but I think we should go one more round.

Keep: 9.2.8's rejection condition (the current text binds to a single-use nonce but never says what happens when details change or the approval comes back twice), 9.2.9 (turns "isolated" into something testable), and 9.2.5's runtime enforcement, which matches what #1141 did for 9.1.1 and 9.1.2.

Restructure: 9.2.2's "matches what will run" is the actual attack, but it's a separate condition - better as a new 9.2.12 than folded into "canonicalized and complete".

Drop:
9.2.10 and 9.2.11 - "highest-impact reversibility classification present anywhere in the chain" is a max() over labels you can ask for; "strictest controls" isn't checkable. Both also drop "action chain" and "cumulative chain", which #1133 added to the glossary one commit ago so these could stay short.
"as applicable" in 9.2.2 - makes a fixed evidence list negotiable.
9.2.7 - narrows to indirect injection, and the trailing definition conflicts with Appendix A having both entries.

The Appendix B edit - AD.19 is headed "Require human approval for high-impact actions", so dropping "high-impact" goes the wrong way. The 9.2.6, 9.2.7, 9.2.10 and 9.2.11 rows also still carry the old wording.

Separately: canonicalization, nonce, execution context and deterministic policy gate belong in Appendix A rather than inline - canonicalized is also in C2.1.2 and nonce in C10.4.6.

Suggest splitting into the 9.2.5 / 9.2.8 / 9.2.9 hardening (as material changes, not readability), a glossary PR, and leaving 9.2.10 and 9.2.11 alone.

ottosulin:
I think the changes look good, more readable in the proposed form

Mayur021:
Agreeing with @RicoKomenda on 9.2.10 and 9.2.11. Declaring the interest first: 9.2.11 is my wording, carried over from #1123, so I am not a neutral party on it.

#1123 was about exactly this. "The chain" could not be assessed because two honest readings gave opposite verdicts on the same system. What landed folds over the reversibility classification of 9.2.3, which names one dimension and orders it, from read-only through to irreversible. "The strictest controls required by any action's impact and reversibility" names no dimension and no ordering anywhere in the document, so an assessor is back to settling what the requirement means before testing it. Same defect as the issue, arriving through the vocabulary instead of the scope.

One thing not yet raised. Appendix B drifts on four rows, not one. It still reads "AI-augmented review" for C9.2.6 and C9.2.7 where the rewrite says "AI-assisted", and still reads "action chains" and "highest-impact reversibility classification" for C9.2.10 and C9.2.11. The diff touches only the 9.2.1 summary, so after merge the inventory and the requirements would disagree on all four. Rico's point about the Action Chain entry that #1133 added to Appendix A is the same problem one file over.

Worth noting that the stated validation compared requirement identifiers and levels. Those are exactly what a vocabulary rewrite preserves, so it would not have surfaced any of this.

The readability intent is right, and 9.2.6 and 9.2.7 do read better in the new form.

jmanico:
Closing this in favor of Rico's split. Thanks @RicoKomenda, @ottosulin and @Mayur021 for the careful read.

Next steps, as separate PRs:

1. **Material changes to 9.2.5, 9.2.8 and 9.2.9**, one sentence each, no inline definitions:
   - 9.2.5: the runtime blocks self-modification outside the defined limits, same pattern as #1141 for 9.1.1 and 9.1.2.
   - 9.2.8: approvals are rejected if a bound value changes or the approval is replayed.
   - 9.2.9: the runtime can neither read the approval keys nor use them to issue approvals.

   Appendix B rows updated to match.
2. **Glossary:** add canonicalization, nonce, execution context and deterministic policy gate to Appendix A. That also covers 2.1.2 and 10.4.6.
3. **Everything else in C9.2 stays as is**, including 9.2.10 and 9.2.11.

Rico's "matches what will run" point for 9.2.2 can be its own issue as a new 9.2.12.

One loose end: #1155 already changed the 9.2.1 row in Appendix B to "privileged or irreversible" so it matches the requirement text. Rico, if you want "high-impact" back there, I'll restore it.

PR #1102, opened 2026-07-13
Adding 10.4.9 for verifying that a MCP proxy server not is vulnerable to Confused Deputy Vulnerability and sharpen 10.2.7 to prevent blocking legitimate use of specifc token flows (#1100)

sydseter:
Closes #1100

## Summary

This PR updates 10.2.7 and adds 10.4.9 to address two different problems that were conflated in the discussion around 10.2.7:

1. token passthrough / token audience misuse
2. OAuth proxy confused-deputy authorization inheritance

The result is a narrower and more accurate 10.2.7, plus a new requirement that directly addresses the shared-upstream-client confused-deputy case.

## Changes

In [1.01-dev/en/0x10-C10-MCP-Security.md](1.01-dev/en/0x10-C10-MCP-Security.md):

- 10.2.7 is changed from:
  - "Verify that MCP servers do not pass through access tokens received from clients to downstream APIs."
- to:
  - "Verify that MCP servers only accept tokens explicitly issued for them."

- 10.4.9 is added:
  - "Verify that MCP proxy servers using a shared upstream OAuth client identity do not allow a requesting MCP client to inherit authorization established for a different MCP client."

## Background

Issue #1100 correctly pointed out that the previous 10.2.7 wording was too broad and could be read as blocking legitimate delegated token flows, while still not clearly expressing the actual MCP requirement.

The MCP authorization and security guidance distinguishes these concerns:

- For token passthrough, MCP guidance says servers must only accept tokens explicitly issued for the MCP server.
- For confused deputy in proxy architectures, MCP guidance focuses on preventing one MCP client from inheriting authorization established for another client when a shared upstream OAuth client identity is in use.

This PR aligns AISVS with that distinction.

## Rationale

The revised 10.2.7 keeps the requirement centred on the verifiable security property that matters: the MCP server must reject tokens not explicitly issued for it. That preserves legitimate architectures where downstream access may use separate, properly issued tokens, without weakening the control against audience confusion and token misuse.

The new 10.4.9 captures the MCP proxy confused-deputy condition at the AISVS level without over-specifying one implementation pattern. It stays testable: two different requesting MCP clients should not be able to share or inherit upstream authorization state.

## Scope

This PR is intentionally limited to the 1.01-dev. It does not attempt to duplicate all lower-level OAuth consent, redirect URI, CSRF, or state-handling details already covered by MCP documentation and related standards.

## References

- MCP Authorization Specification:
  https://modelcontextprotocol.io/specification/2025-06-18/basic/authorization
[...]

Issue #1100, opened 2026-07-02
10.2.7 is misunderstood and does not help to prevent the confused deputies Supabase/Cursor Incident


sydseter:
The following requirement seems to have been created based on some misunderstandings or rewritten to the point that it no longer holds any authority as a proper requirement.

10.2.7 | Verify that MCP servers do not pass through access tokens received from clients to downstream APIs. | 2
-- | -- | --

The rationale for the requirement is based on the research as it is documented within the research chapters: https://github.com/OWASP/AISVS/blob/main/1.0/research/chapters/C10-MCP-Security/C10-02-Authentication-Authorization.md

10.2.7 | Verify that MCP servers do not pass through access tokens received from clients to downstream APIs. | 2 | Horizontal privilege escalation via argument manipulation. Supabase Cursor agent incident (mid-2025) — agent ran with privileged service-role access and processed user-supplied SQL in support tickets, exfiltrating integration tokens. No standard framework exists for argument-level authorization in MCP. | Test tool invocations with arguments outside the user's authorized scope (e.g., different tenant IDs, unauthorized file paths). Verify the server rejects based on token claims, not just tool name. Review authorization middleware for argument inspection. Traefik TBAC's Transaction authorization enforces parameter-level constraints. AWS Bedrock AgentCore Gateway interceptors can enforce tool-level and argument-level policies at the gateway, contextually filtering ListTools, InvokeTool, and Search calls based on the calling principal's permissions. | Argument-level authorization is significantly harder than tool-level. Requires the server to understand argument semantics. As of April 2026, four products address this: Traefik TBAC (Transaction authorization), AWS AgentCore Gateway (interceptors), Oso (Polar policy-as-code supporting RBAC/ReBAC/ABAC simultaneously), and Kong Tool ACLs (tool-level only). Oso evaluates per-token identity against declarative policies, centralizing authorization logic outside MCP server code. Gravitee MCP Proxy adds OpenFGA/AuthZen for relationship-based authorization. AuthZed/SpiceDB and OPA/Rego do not yet have production MCP integrations (SpiceDB ships an MCP server for development, but not as a policy engine for MCP tool authorization).
-- | -- | -- | -- | -- | --

The problem with this is the following:

If you categorically enforce that "tokens received from clients must not be passed to downstream APIs," you prevent legitimate, secure agent architectures—such as 3-legged OAuth 2.0 delegated flows and specific identity propagation methods needed to maintain audit logs.
[...]

PR #1158, opened 2026-09-20
MCP protocol 2026-07-28 version changes

ottosulin:
# C10 MCP Security: update for MCP specification 2026-07-28

Revises `1.01-dev/en/0x10-C10-MCP-Security.md` to match the published MCP specification revision 2026-07-28. 

Five existing requirements changed, fourteen new requirements added, seven candidates deferred. Deferred candidates are included here for completeness and for future reference.

The 2026-07-28 revision removes the `initialize` handshake and protocol sessions, moves version and capability negotiation into per-request `_meta`, mirrors tool parameters into HTTP headers, and adds response caching and URL-mode elicitation. Requirements written against the session model no longer describe testable behaviour.

Every requirement below is grounded in normative text from the published specification. Conditional MUSTs are worded with their condition. 

Requirement text column describes the proposed requirements (yes you could read this from the changes, but here for readability) and Rationale column has short description about the change.

## Changes to existing requirements

| # | Requirement text | Rationale |
|---|---|---|
| **10.2.4** | **Verify that** MCP tools/list returns only tools permitted by the requester's authorized scopes, and that list responses vary only by authorization context, never by connection. | List contents vary by authorization context only. Precondition for safe caching: `private` responses must not be shared across authorization contexts. Level 2 unchanged. |
| **10.2.6** | **Requirement removed** | Not in the protocol anymore. |
| **10.2.7** | **Verify that** MCP servers only accept tokens explicitly issued for them and neither accept nor transit tokens issued for other services. | Backed by spec MUSTs: RFC 8707 resource parameter, audience validation (= prohibition to transit tokens), token passthrough prohibition. Consistent with 10.2.1 and 10.2.2. **Level 2 to Level 1.** |
| **10.3.3** | **Verify that** MCP servers validate both the Origin header and the Host header independently on all HTTP-based transports to prevent DNS rebinding attacks. | Text unchanged. Origin validation is an unconditional MUST with trivial implementability. **Level 2 to Level 1.** |
| **10.3.4** | **Verify that** MCP clients enforce a minimum acceptable protocol version and do not retry with, or fall back to, a protocol version or legacy handshake below that minimum. | Responses no longer carry a protocol version, so the previous text tests nothing. The specification documents the downgrade path explicitly: a modern request returning `4xx` without a recognized modern error body causes the client to fall back to `initialize` and possibly further to the deprecated HTTP+SSE transport. Level 2 retained during the interop window. |

## New requirements

| # | Requirement text | Rationale |
|---|---|---|
| **10.2.8** | **Verify that** MCP servers treat request state received from clients as attacker-controlled input, and that where such state influences authorization, resource access, or business logic, servers protect its integrity, reject state that fails verification, and enforce single-use server-side where the state must be consumed at most once. | Treating the value as attacker-controlled is an unconditional MUST. Integrity protection is a MUST conditioned on the state influencing authorization, resource access or business logic, and may be omitted only where tampering can cause nothing worse than request failure. Single-use enforcement is a MUST for servers whose state must be consumed at most once, and the spec states the SHOULD bindings (principal, TTL, originating request) do not achieve it. Those bindings are SHOULD and are not asserted here. Level 1. |
| **10.2.9** | **Verify that** MCP clients persisting OAuth client credentials associate those credentials with the authorization server that issued them, and re-register with the new authorization server rather than reusing credentials when the resource's authorization server changes. | Spec MUST NOT plus MUST: clients must not reuse client credentials from a different authorization server and must re-register. Credential reuse across authorization servers exposes client authentication material to an unintended party. Level 2. |
| **10.3.6** | **Verify that** MCP intermediaries enforcing policy on mirrored request headers confirm that the protocol version indicates a revision requiring header and body validation before trusting those headers. | Spec SHOULD. Revisions that predate header and body validation do not guarantee the headers match the body, so a gateway routing or rate-limiting on mirrored headers can be steered by a header that the origin never validates. Level 2. |
| **10.4.11** | **Verify that** MCP clients reject tool definitions containing invalid mirrored-header annotations and exclude only the affected tool from the tool list. | Spec MUST. Annotation names must match HTTP field-name token syntax and must not contain CR or LF, otherwise header injection is possible through a tool definition. Rejection is scoped to the individual tool so one malformed definition does not disable the rest. Client-side, evaluated at tool-list time. Level 1. |
| **10.4.12** | **Verify that** MCP servers reject requests where mirrored request headers do not match the corresponding request body values, after decoding any encoded header values. | Spec MUST, rejected with `400 Bad Request` and JSON-RPC error `-32020` (`HeaderMismatch`). Servers must decode sentinel-encoded values before comparison, otherwise a legitimately encoded value is read as a mismatch and an attacker-chosen encoding evades the comparison. Server-side, evaluated per request, distinct from 10.4.11. Level 2. |
| **10.4.13** | **Verify that** MCP implementations apply resource bounds to schema validation, such as a maximum schema depth, a cap on the total number of nested schema elements, or a per-validation time budget. | Spec SHOULD. Composition keywords and `$defs` allow a hostile schema to act as a denial-of-service vector against the validator. Binds implementations, typically whichever party validates a hostile schema. Split from 10.4.5. Level 2. |
| **10.4.14** | **Verify that** MCP implementations do not automatically dereference JSON Schema `$ref` values that resolve to network URIs, and that any opt-in external resolution is disabled by default and rejects loopback, link-local, and private network addresses. | Spec MUST NOT and guardrails. Server-side request forgery attack vector, reachable through a tool or resource schema. Level 1. |
| **10.4.15** | **Verify that** MCP implementations reject schemas that fail to validate due to an unresolved external `$ref` rather than treating them as permissive. | Spec SHOULD. Failing open silently voids 10.4.1, 10.4.3 and 10.4.4: an unresolvable reference causes the schema to admit input it was written to reject. Level 2. |
| **10.4.16** | **Verify that** MCP clients and gateways do not serve cached responses across authorization contexts and do not cache results of multi round-trip requests. | Spec MUSTs. Caches must not be shared across authorization contexts, and requests carrying `inputResponses` or `requestState` must not be cached because they depend on inputs outside the cache key. Cross-context reuse is a new avenue for leaks. `ttlMs` is a freshness hint and is excluded. Level 2. |
| **10.4.17** | **Verify that** MCP caches are not shared across authorization contexts, and that responses marked with a private cache scope are reused only within the authorization context that produced them. | Spec MUST: caches MUST NOT be shared across authorization contexts, a different access token requires a different cache. Cross-context reuse of a cached response served from an authenticated endpoint leaks data between callers. Correct labelling of cache scope is a SHOULD and the spec states servers MUST NOT rely on cache scope alone for access control, so the requirement targets the sharing boundary rather than the label. Level 1. |
| **10.4.18** | **Verify that** MCP servers do not use form-mode elicitation to request secrets or payment credentials, and that in URL-mode elicitation the user completing the flow is the same user who initiated it. | Both are server-side MUSTs. The same-user binding prevents an account-takeover flow where an attacker induces a victim to complete an authorization the attacker initiated, binding the victim's third-party tokens to the attacker's identity. Level 2. |
| **10.4.19** | **Verify that** MCP servers exclude end-user credentials and personal data from URL-mode elicitation URLs, and do not issue URLs that are pre-authenticated to a protected resource. | Two server-side MUST NOTs. A pre-authenticated URL passing through the client is directly usable to impersonate the user. Level 1. |
| **10.4.20** | **Verify that** MCP clients require explicit user consent and display the full URL before opening any URL-mode elicitation target, and do not pre-fetch the URL or its metadata. | Client-side MUSTs. Pre-fetching is a separate obligation from consent: it discloses the URL to the client before the user has approved anything. Distinct verification case from the server-side elicitation requirements. Level 2. |
| **10.4.21** | **Verify that** MCP clients open URL-mode elicitation targets so that neither the client nor the model can observe the page contents or the user's input. | Client-side MUST. This is the purpose of URL mode: the flow exists so credentials and third-party session content stay outside the model's context. An embedded webview satisfies consent and URL display while still exposing the page to the client. Level 1. |

## Deferred requirements

| # | Requirement text | Rationale |
|---|---|---|
| n/a | Clients validate the `iss` parameter on authorization responses against the issuer recorded before the redirect. | Covered by ASVS 5.0 10.2.2 (mix-up attack defence via `iss`). Removing the MCP noun leaves a general OAuth client requirement. The MCP comparison rules (no URI normalization, applies to error responses) are implementation nuance for satisfying that requirement, not a separate control. Chapter References to note the ASVS dependency maybe? |
| n/a | Clients implement PKCE with S256 and refuse to proceed without code challenge support. | Covered by ASVS 5.0 10.2.1 (client) and 10.4.6 (authorization server, including rejection of `plain`). No MCP-specific delta. |
| n/a | Servers generate state handles with secure randomness and bounded lifetime. | No normative backing. The specification states the protocol has no concept of a state handle and marks the containing section non-normative. The entropy guidance applies only to unauthenticated servers where the handle is necessarily a bearer token. |
| n/a | Servers do not advertise `offline_access` and do not assume refresh tokens will be issued. | Server obligation is SHOULD NOT. Generic OAuth behaviour with no AI-specific nuance. |
| n/a | Clients compute the union of granted and challenged scopes during step-up authorization. | Client obligation is SHOULD within a SHOULD flow. Generic OAuth 2.1 behaviour. |
| n/a | Implementations do not base security decisions on self-reported `_meta` fields, and sanitize telemetry keys before forwarding. | Self-reported fields are SHOULD NOT, not MUST NOT. The telemetry sanitization half has no specification backing. Compound as written. Reconsider for 2.0. |
| n/a | MCP App UI resources render sandboxed with a default CSP, and UI-initiated tool calls traverse the same authorization, consent, and audit path as model-initiated calls. | MCP Apps is an independently versioned extension. Core chapter requirements should not depend on extension behaviour. Deferred to 2.0, together with Tasks, Enterprise-Managed Authorization (SEP-990) and Skills over MCP (SEP-2640, no published normative specification). |


1.01-dev/en/0x10-C10-MCP-Security.md
```diff
@@ -27,10 +27,12 @@ Callers must be authenticated and access to MCP servers authorized, following pr
...
+| **10.2.4** | **Verify that** MCP tools/list returns only tools permitted by the requester's authorized scopes, and that list responses vary only by authorization context, never by connection. | 2 |
 | **10.2.5** | **Verify that** MCP servers enforce access control on every tool invocation, validating that the user's access token authorizes both the requested tool and the specific argument values supplied. | 2 |
-| **10.2.6** | **Verify that** MCP servers ensure all session artifacts are removed when a session terminates. | 2 |
-| **10.2.7** | **Verify that** MCP servers only accept tokens explicitly issued for them. | 2 |
+| **10.2.6** | Requirement removed. | |
+| **10.2.7** | **Verify that** MCP servers only accept tokens explicitly issued for them and neither accept nor transit tokens issued for other services. | 1 |
+| **10.2.8** | **Verify that** MCP servers treat request state received from clients as attacker-controlled input, and that where such state influences authorization, resource access, or business logic, servers protect its integrity, reject state that fails verification, and enforce single-use server-side where the state must be consumed at most once. | 1 |
```
RicoKomenda, line 34: I know why this requirement exists, but I think it's a tad long. Maybe we'll find another shorter version of it?

Maybe split it into 2 requirements for e.g.:

`| **10.2.8** | **Verify that** MCP servers protect the integrity of request state that influences authorization, resource access, or business logic, and reject any such state that fails verification. | 1 |

| **10.2.9** | **Verify that** MCP servers enforce single-use, server-side, for request state that must be consumed at most once. | 1 |`

RicoKomenda (commented):
One requirement I'm not 100% happy with. The others ones are superb, great job @ottosulin

jmanico:
Thanks @ottosulin, and +1 to Rico. This is careful work. I checked everything against the published 2026-07-28 text. Numbering follows the 1.01 rule (10.2.6 retired in place, new IDs appended), and the suggestions below keep it.

**Before merge**

1. **10.4.16 / 10.4.17 (caching).** The spec's "caches MUST NOT be shared across authorization contexts" applies to `private` only. For `public`, any client, gateway, or proxy MAY serve the cached response to any user ([Cache Scope Field](https://modelcontextprotocol.io/specification/2026-07-28/server/utilities/caching#cache-scope-field)). As written, both would fail a spec-compliant gateway, and they duplicate each other. Keep both IDs and split them by party:
   - **10.4.16** Verify that MCP clients and intermediaries reuse cached responses marked with a private cache scope only within the authorization context that produced them, and do not cache results of multi round-trip requests. **L2**
   - **10.4.17** Verify that MCP servers mark cacheable results containing user-specific or authorization-filtered data with a private cache scope. **L2**

   Server mislabeling is the bigger leak (#823), and it's SHOULD-level like 10.3.6, 10.4.13 and 10.4.15. Both are L2 because the leak needs a shared cache.

2. **10.2.8.** +1 to Rico's split. Name `requestState` explicitly, and append the second half as 10.2.10 because 10.2.9 is taken. I'd put single-use at L2: it needs a replay store shared across instances, which [MRTR](https://modelcontextprotocol.io/specification/2026-07-28/basic/patterns/mrtr) is designed to avoid, and it only applies to consume-once state.
   - **10.2.8** Verify that MCP servers protect the integrity of `requestState` that influences authorization, resource access, or business logic, and reject state that fails verification. **L1**
   - **10.2.10** Verify that MCP servers enforce single use server-side for `requestState` that must be consumed at most once. **L2**

3. **10.4.21:** The description says L1 and the file says L2. L2 matches 10.4.20.

4. **Appendix B** still lists C10.2.6 and has none of the new IDs. That can go here or in a follow-up before 1.01.

**Smaller**

- **10.2.7:** "accept" appears twice. Suggest "…and do not pass tokens issued for other services through to downstream APIs." That's narrower than the text #1102 removed, so #1100's delegated flows are unaffected. cc @sydseter
- **10.2.9:** Re-registration doesn't fit pre-registered clients, which the spec says should surface an error instead ([AS binding](https://modelcontextprotocol.io/specification/2026-07-28/basic/authorization/client-registration#authorization-server-binding)). Suggest "Verify that MCP clients bind persisted OAuth client credentials to the issuing authorization server and do not present them to a different authorization server."
- **10.4.11:** Drop "and exclude only the affected tool from the tool list". That's about availability, not security.

Everything else matches the spec, including the deferrals.

Opus 5.5 pass, grain of salt as usual. Override anything you disagree with.

jmanico (commented):
comments added below

